PR #385, opened 2014-04-15
Added new option: showActiveTooltip

kylerberry:
Added an option to persistently show the tooltip of the active section.

### DIFF
--- a/jquery.fullPage.js
+++ b/jquery.fullPage.js
@@ -874,7 +874,7 @@
 			mouseenter: function(){
 				var tooltip = $(this).data('tooltip');
 				
-				if(options.showActiveTooltip === true) {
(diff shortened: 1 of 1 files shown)


kylerberry:
Can we talk about adding this? I know it needs refactoring at this point, which I'd be happy to do.

alvarotrigo:
@kylerberry yep, we can. 
Probably it would be much easier now, you probably would only need to deal with CSS now, as the tooltip content is static at the moment.

kylerberry:
@alvarotrigo Thanks, I'll make a PR soon

kylerberry:
@alvarotrigo Closing this for a fresh PR since this is so old. See new PR: https://github.com/alvarotrigo/fullPage.js/pull/1081

PR #477, opened 2014-05-25
renamed bower dependency "jQuery" => "jquery"; added bower_components to...

theborakompanioni:
the bower.json file in the downloaded package states "jquery" as the name.
it is important because it impacts scripts with further libs depending on "jquery" .
e.g. when automatically generating sources => jquery gets injected twice.

``` html
<!-- build:js scripts/vendor.js -->
<!-- bower:js -->
<script src="bower_components/jquery/dist/jquery.js"></script>
<script src="bower_components/jQuery/dist/jquery.js"></script>
<!-- endbower -->
<!-- endbuild -->
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 .DS_Store
+bower_components/
--- a/bower.json
+++ b/bower.json
@@ -27,6 +27,6 @@
   ],
   "dependencies": {
       "jquery.slimscroll": "latest",
-      "jQuery": "latest"
+      "jquery": "latest"
     }
 }


alvarotrigo:
@theborakompanioni the .gitignore is not needed as far as I understand. Is it?

theborakompanioni:
@alvarotrigo Hmm.. it seems to me, that you are not using bower locally, are you? Otherwise the  "bower_components" directory would have been in your repo. I'd suggest to leave it in .gitignore (although others add their bower dependencies into the repo - this is because some library maintainers are lazy sometimes and change already published modules [like it happened with jQuery!]). 

Nonetheless - it is useful if you are thinking about adding tests and/or let a ci-server build your script.

alvarotrigo:
Yeah ok. Thanks for the info!
And yeah, it would be nice to add some tests in a future...

theborakompanioni:
@alvarotrigo Having tests in place will strengthen the confidence in the changes committed to the repo - I certainly do not want to develop without them anymore.

It is sometimes relatively tricky to test libraries which heavily manipulate the DOM. I'll happily invite you to take a look at [VisSense.js](https://github.com/vissense/vissense) which runs test with [jasmine](https://jasmine.github.io/) and [Karma](http://karma-runner.github.io) - it has a coverage of 100% (although not all conditional branches) and it will not build if the coverage falls below 95%!. 

Of course there are other great javascript testing frameworks out there.

alvarotrigo:
I definitely would take a look at them and hopefully I can find some time to start with the testing.

PR #613, opened 2014-08-03
added slimScroll event documentation

Kwaadpepper:


alvarotrigo:
It was replaced by iScroll.js

PR #786, opened 2014-10-20
Fix mobile after resize

sugarshin:
can be obtained correctly

### DIFF
--- a/jquery.fullPage.js
+++ b/jquery.fullPage.js
@@ -204,8 +204,8 @@
 		$.fn.fullpage.reBuild = function(resizing){
 			isResizing = true;
 
-			var windowsWidth = $(window).width();
-			windowsHeight = $(window).height();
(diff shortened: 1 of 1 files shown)


kimek:
Hi Alvarotrigo, as above i don't know issue about which sugarshin mentioning here. I have small pull request related to resize. Could You add paddingTop/paddingBottom to $.fn.fullpage.reBuild() or to Methods as on resizing I have (and probably other people too) problems with setting other padding for mobile/desktop version. I know that i can use destroy and implement fullpage.js all the time when somebody rotate the tablet (or even phone as their are quite big now-a-days). Thanks. Awesome Job. Warmest regards !

alvarotrigo:
@kimek you can probably change that using javascript or jQuery on the `afterReBuild` callback. No need to change the plugin at all.

PR #814, opened 2014-11-06
Vertical navigation arrows

jacobmllr95:
- Add option to show arrows to move to the previous/next section on the
  vertical navigation
- Add a demo
- Some cleanup
- Bump version to 2.4.4


okadots:
4 years later...

@jackmu95 i used @alvarotrigo's solution above, but added the mark up for the arrows inside the navigation div with js. i then positioned them absolutely, relative the navigation wrapper

`afterRender: function(){
			$('#menu').append('<span class="arrow arrowUp"></span><span class="arrow arrowDown"></span>');
	}`

PR #858, opened 2014-11-20
slimScroll wheelstep configuration

jefsnare:
According slimScroll documentation (http://rocha.la/jQuery-slimScroll) it's possible to configure the wheelstep option (mouse wheel data).


jefsnare:
Not important any more, for now I 'cancel' this pull request

PR #971, opened 2015-01-26
exception handling

YuyingWu:
If jQuery lib do not exist, tell the user in console: Error:
fullPage.js requires jQuery library. (1.6.0 minimum)


alvarotrigo:
Thanks for the pull request!
But it says: 

> 1,581 additions, 1,577 deletions not shown

Please solve that. I can not accept a pull request like that.

PR #982, opened 2015-02-01
Update to jquery.fullPage.css: Solving blinking(hiding) arrows or slides navigation when scrolling -  solved by alexeyenko

gtluszcz:


alvarotrigo:
You should also add the other prefixes:
```css
  -webkit-transform: translate3d(0,0,0); 
      -ms-transform: translate3d(0,0,0);
          transform: translate3d(0,0,0);
```

gtluszcz:
yes you're right thank you

PR #1014, opened 2015-02-11
avoid TypeError

tgiunipero:
We're getting a type error "Cannot read property 'top' of undefined" - I'm adding a condition to check if an element exists before calling position().  Note, this is already done above within the method [here](https://github.com/alvarotrigo/fullPage.js/blob/master/jquery.fullPage.js#L99).

Error occurs when destroy('all') is called.


alvarotrigo:
Thanks for it!

PR #1030, opened 2015-02-18
optimized use of variables

tudor-sv:
Optimized use of variables. Freed  up 1kb :)


alvarotrigo:
Thanks for it. 
The options variable is only present 3 times in the current minified version, so there's no need to change it.

If you can commit another pull request with only the changes in the other variables, that'd be nice.

alvarotrigo:
Oh wait, now that I think about it. We only need the `$.fn.fullpage` substitution. 

The other can be achieved like this:
```javascript
(function($, window, document) {
    $.fn.fullpage = function(options) {
    };
})(jQuery, window, document);
```

tudor-sv:
True, great suggestion !
Also added Math as a parameter (it appeared 14 times before in the minified version)

alvarotrigo:
Thanks for it!

tudor-sv:
my pleasure

PR #1037, opened 2015-02-20
Added background-color hack to dectect hover state on page load.

jnettik:
Detects hover over a slimScroll area on pageload. It uses the CSS hack found at http://stackoverflow.com/a/6593995/1547641. I'm including the unminified version of the same slimScroll.js for clearity. My changes start at line 250.

This is discussed in rochal/jQuery-slimScroll#133

### DIFF
--- a/jquery.fullPage.css
+++ b/jquery.fullPage.css
@@ -85,7 +85,9 @@ html, body {
 }
 .fp-scrollable {
     overflow: scroll;
-
+}
+.fp-scrollable:hover {
+    background-color: transparent;
 }
 .fp-notransition {
     -webkit-transition: none !important;
--- a/vendors/jquery.slimscroll.js
+++ b/vendors/jquery.slimscroll.js
@@ -1,0 +1,480 @@
+/*! Copyright (c) 2011 Piotr Rochala (http://rocha.la)
+ * Dual licensed under the MIT (http://www.opensource.org/licenses/mit-license.php)
+ * and GPL (http://www.opensource.org/licenses/gpl-license.php) licenses.
+ *
+ * Version: 1.3.2
+ *
+ */
+(function($) {
+
+  jQuery.fn.extend({
+    slimScroll: function(options) {
+
+      var defaults = {
+
+        // width in pixels of the visible scroll area
+        width : 'auto',
+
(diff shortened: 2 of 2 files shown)


alvarotrigo:
Testing it a bit more it doesn't seem to solve the problem. Still have troubles sometimes.
I'm testing it with srollable content on the 2nd section, and it enters in the condition you created always, not only when its on the 2nd section.

Even when the hover event is not beeing applied as you can see [in this example](http://jsfiddle.net/u9P87/13/) I made with your modification.

jnettik:
What browsers and OS are you testing on? I'm using the modification in a project I'm working on so far without issue. Even the JSFiddle you linked to seems to be working for me.

alvarotrigo:
The JSFiddle I posted shows that the hover is fired even when the mouse is not really over the scrollable area. 
Which means it is always true, and you can even notice the CSS style for :hover is not even in there and still acting the same way.

alvarotrigo:
Using Chrome in Windows 7 and Mac Yosemite.

jnettik:
Ah I see now. Looks like the default value for `background-color` in CSS is `transparent`. That is why this is always true. http://www.w3schools.com/cssref/pr_background-color.asp That throws a wrench in this. I just tested with a more noticeable color and even CSS, at least in Chrome on Yosemite, doesn't detect hover on page load.

Which means the we're back to choosing between not being able to scroll until mouse movement, or not being able to use `overflow: scroll;` on regions outside of slimScroll.js.

jnettik:
Looking into some of the other scrolling JavaScript plugins, they don't seem to have this issue. You've already mentioned you were considering iScroll, which is one I tested that seems to work. It does seem like using a different plugin is the better long term solution to this.

dalla777:
I'm unsure if this issue was solved at one point, but my fix for this was to manually call a mouseover. In the afterSectionLoads function, I simply added `v.element.find('.fp-scrollable').mouseover();` and that seems to have done the trick for me.

alvarotrigo:
@dalla777 great solution! Thanks for sharing it! I'll add it in the next release!

alvarotrigo:
Added in the latest version 2.6.5.

PR #1050, opened 2015-02-27
Bootstrap responsive visibility

Guilain:
Add :visible filter to move up & move down functions


alvarotrigo:
That might carry other problems in other parts of the code...

Guilain:
Like moveTo function ?

alvarotrigo:
Yeah, for example.

Also, why is it related with bootstrap? Is it because bootstrap sets the visibility of a section to `hidden` when resizing to certain point?

I would have to review it with more time, but it can carry some problems, such as:

- If you hide the 1st section, it can be set to `active` by default on page load.
- If you hide the active section on resizing. What will happen ?
- If the active section on page load is hidden (when setting an active section/slide other than the 1st), what will happen?
- Possible problems when using `autoScrolling:false` as per [this](https://github.com/alvarotrigo/fullPage.js/blob/master/jquery.fullPage.js#L577).
- Keyboard navigation for the Home / End keys

Guilain:
I've updated to your 2.5.9
Some problems are fixed, but it's not ready yet ...

alvarotrigo:
@Guilain you've done something wrong. It says "2,081 additions, 2,038 deletions not shown" in your last commit.

PR #1057, opened 2015-03-03
function afterSectionLoads callback time

thebeautyoflove:
Maybe the time of function afterSectionLoads should be defined by users


alvarotrigo:
Yeah, or maybe just remove it completely.

alvarotrigo:
The delay was completely removed time ago.

PR #1086, opened 2015-03-16
Fix setBodyClass method

mgmartel:
When I updated fullPage.js I noticed that the body class had changed to `fp-viewing0` instead of `fp-viewing-0` (this started happening in rev 9702c05). Added the dash back in.

Also updated the regexp that cleans the body classes to use the constant `VIEWING_PREFIX`.

Thanks for this awesome script! Keep up the good work :+1:


alvarotrigo:
Thanks!!

PR #1087, opened 2015-03-16
Explicit padding-top/bottom inline styles

pmhou:
Only apply top and bottom padding values so not to overwrite existing left/right padding values of the section element.


alvarotrigo:
Thanks for it!

PR #1092, opened 2015-03-18
Bower version errors resolve.

usmantahirr:
I was getting jQuery version error again and again without any specific solution. So I decided to for it and change this.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -29,6 +29,6 @@
     "examples"
   ],
   "dependencies": {
-      "jquery": "2.x"
+      "jquery": "^2.x.x"
     }
 }
--- a/bower.json
+++ b/bower.json
@@ -1,6 +1,6 @@
 {
   "name": "fullpage.js",
-  "version": "2.6.2",
+  "version": "2.6.3",
   "homepage": "http://alvarotrigo.com/fullPage/",
   "authors": [
     "Alvaro Trigo https://github.com/alvarotrigo"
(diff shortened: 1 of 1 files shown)


alvarotrigo:
I would merge this change `"jquery": "^1.x"` but not the one with jquery 2.X.
Remember that fullPage.js gives support to IE 8.

usmantahirr:
But this creates problem when I use it with foundation or any other plugin which requires JQuery 2.x
because it throws conflict error while doing bower install

alvarotrigo:
I guess than then we should be using `>=1.6.0` then.

usmantahirr:
I will test it and if it works, I will send a pull request again. If it works fine at my side

alvarotrigo:
Any news?

alvarotrigo:
@usmantahirr did you tried it?

usmantahirr:
Not yet. I will do it today and send a pull request again.

alvarotrigo:
@usmantahirr what about now?
Otherwise I will just commit the changes myself.

usmantahirr:
I am not getting any error now. Its now asking me for a JQuery version. I was getting that error on My OfficePC. Its Windows and have an older version of Node and NPM. I dont know Older versions have such issue. But now I guess this pull request isn't required.

![Image of Yaktocat]
(http://i.imgur.com/ymAXkLt.png)

PR #1095, opened 2015-03-20
removed old prefixes

tomByrer:
Sorry must have missed a few https://github.com/alvarotrigo/fullPage.js/pull/579
Maybe remove the [`box-sizing` prefixes](http://caniuse.com/#feat=css3-boxsizing) also?

### DIFF
--- a/jquery.fullPage.css
+++ b/jquery.fullPage.css
@@ -164,8 +164,6 @@ html, body {
     top: 50%;
     margin: -2px 0 0 -2px;
     -webkit-transition: all 0.1s ease-in-out;
-    -moz-transition: all 0.1s ease-in-out;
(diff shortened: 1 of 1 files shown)


alvarotrigo:
I'm not quite sure until which version of Opera or Mozilla those vendors are needed. Do you have any?

tomByrer:
[transitions](http://beta.caniuse.com/#search=transitions) or box-sizing?
(On CanIUse, click on [Show All] button above the browser list to see older versions)

alvarotrigo:
@tomByrer yeah, but that is not showing older versions of other browsers such as Opera 12 and so on. 
I guess, that the best thing is to keep them just in case, as they wont harm anybody :)

Issue #1148, opened 2015-04-15
Slide transition causes blurry text


rudyryk:
Hi,
the fullPage.js plugin as great!

I've just noticed that while performing slide transition the plugin doesn't round transition values:
![screen shot 2015-04-15 at 19 08 51](https://cloud.githubusercontent.com/assets/4500/7163541/471b7c72-e3a3-11e4-95fc-6b10ad8e64aa.png)

- that's a screenshot for completed slide animation. This may cause **blurry text** on slide, so does it in my case.

I'm not sure, if we should round transition coordinates before run, but I believe it would be right at least to have rounded values after animation completed.

alvarotrigo:
Have you tried using `-webkit-perspective: 1000;`. 
Sounds similar [to other problems caused by translate3d](http://stackoverflow.com/questions/8024061/webkit-blurry-text-with-css-scale-translate3d).

rudyryk:
I'm using slides inside page grid, particulary in Semantic-UI grid within column sized 13.0/16.0 of full window width, so internal column width value could be fractional. I've edited div's transition property in debug console by hand and the in slide text became sharp again :)

rudyryk:
BTW text blurring doesn't appear on vertical scrolling, because sections heights are always equal to window height, which is rounded number.

alvarotrigo:
Have you tried using `-webkit-perspective: 1000;`?

rudyryk:
Álvaro, just tried, didn't help. Also in discussion at Stack Overflow mentioned that any transition causes blur, and this is the different case. 

It seems to me that we can just replace `destinyPos.left` to `Math.round(destinyPos.left)` in `landscapeScroll()` when setting `translate3d ` and `scrollLeft`.

rudyryk:
That worked :) It seems that this change is unobtrusive and won't affect other logic, but I'm sure for 99%, not 100%.

alvarotrigo:
No problem, and thanks for the pull request, I'll merge it soon!

PR #1149, opened 2015-04-15
Use rounded values for horizontal scroll transformations

rudyryk:
Related discussion:
https://github.com/alvarotrigo/fullPage.js/issues/1148


alvarotrigo:
Thanks for it!

PR #1166, opened 2015-04-22
scrollOverflow improvements

d13:
- Using native scrolling on touch devices to preserve momentum scrolling
- Debouncing MouseWheelHandler when scrolling inside a section to prevent changing sections too quickly
- Hiding slimscroll-bars when switching slides which prevents the bar from sliding across the screen


alvarotrigo:
Thanks for it @d13 ! I'll try it whenever I have time!
Could you please commit again the changes in the CSS file again? Right now it seems indentation got changed or something similar and I can not clearly see the changes because all lines are reported as new.

d13:
Well do, not sure why it did that.

alvarotrigo:
The only problem I see on this is that it is adding much code into fullPage.js for an optional functionality as the scrollOverflow.

It would be better if it could be added directly into the forked slimScroll.js

PR #1175, opened 2015-04-23
Fixes Chrome vertical height issue.

Synchronizing:
Seems to fix issue relating to Chrome's inability to pull correct height. Haven't tested within other browsers.

Before:
Before

After:
After


alvarotrigo:
Try to reproduce that issue somewhere where I can see it. 
The solution you propose doesn't seem to be a good fix as you are replacing pixels for percentages instead of finding out why the measure in pixels is returned wrong.

PR #1193, opened 2015-05-01
 New arrows for the sliders 

Farkal:
Because you need beautiful arrows for your slider :smile: 
I have just design new arrows, for me they are more funny than the previous. If you think the same feel free to pull :)
Have a nice day.

### DIFF
--- a/jquery.fullPage.css
+++ b/jquery.fullPage.css
@@ -1,203 +1,271 @@
 /**
- * fullPage 2.6.4
+ * fullPage 2.6.0
  * https://github.com/alvarotrigo/fullPage.js
  * MIT licensed
  *
  * Copyright (C) 2013 alvarotrigo.com - A project by Alvaro Trigo
+ * Contribution by Jhony0311
  */
-html.fp-enabled,
-.fp-enabled body {
-    margin: 0;
-    padding: 0;
-    overflow:hidden;
-
-    /*Avoid flicker on slides transitions for mobile phones #336 */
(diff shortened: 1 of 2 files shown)


alvarotrigo:
They are not working on IE 8.

Farkal:
You already use transition and they are not supposed to work before IE 10 -> http://www.w3schools.com/css/css3_transitions.asp
So i don't think it's a problem.
For transform it's the same -> http://www.w3schools.com/cssref/css3_pr_transform.asp
And you already use transform and transition in your own code :/

alvarotrigo:
@Farkal that's not quite true. fullPage.js gives support to IE8 by automatically falling back to jQuery animations when the browser doesn't support css3.

If you read the docs you can see the following:

> **Compatibility**
fullPage.js is fully functional on all modern browsers, as well as some old ones such as Internet Explorer 8, 9, Opera 12, etc. It works with browsers with CSS3 support and with the ones which don't have it, making it ideal for old browsers compatibility.

Farkal:
Yes but there is no way he can use this jQuery animations for my transition and transform as he already do with your's ?

alvarotrigo:
@Farkal the point is: your arrows are not compatible with IE 8 and therefore can not yet be added to the plugin as the default ones.

Farkal:
But do you think it's possible to make them compatible with jQuery animation ?
I can try to adapt them :)

alvarotrigo:
It is not a matter of animations but of CSS.

Farkal:
Ok then can i adapt them with CSS ?

PR #1194, opened 2015-05-01
Added a callback that could be triggered before a slide move.

specialbrand:
Hello,

I had a need for a callback in my project that could be triggered before a slide moves so I could prevent the slide from moving under certain conditions. Feel free to use this minor change for the fullPage.js project as you see fit.

Cheers

### DIFF
--- a/jquery.fullPage.js
+++ b/jquery.fullPage.js
@@ -136,7 +136,8 @@
             afterResize: null,
             afterReBuild: null,
             afterSlideLoad: null,
-            onSlideLeave: null
+            onSlideLeave: null,
+            onBeforeMoveSlide: null
         }, options);
 
         displayWarnings();
@@ -984,6 +985,14 @@
                 destiny = currentSlide.prev(SLIDE_SEL);
             }else{
                 destiny = currentSlide.next(SLIDE_SEL);
+            }
+
+            // before slide move callback
+            if (options.onBeforeMoveSlide && $.isFunction( options.onBeforeMoveSlide ))
+            {
+                if (options.onBeforeMoveSlide.call(this, direction, currentSlide, destiny, slides, activeSection) === false) {
+                    return;
+                }
             }
 
             //isn't there a next slide in the secuence?


faliqaiman:
I would love to the this example so I can implement with my project.

alvarotrigo:
@faliqaiman why dont you use `onLeave`? 
Take into account that you can always [cancel the movement before it takes place](https://github.com/alvarotrigo/fullPage.js#cancelling-the-scroll-before-it-takes-place).

ludovicpages:
Hi all.
Return false on onLeave is great to prevent scrolling.
But i can't find a way to scroll after some delay ( using timeout ).
Something like this, doesn't work.
`onLeave: function(index, nextIndex, direction)
	    {
	    	var timeout = setTimeout(function()
        	{
        		$.fn.fullpage.moveSectionDown();
        	}, 1000);
	    	return false;
        }`

alvarotrigo:
@ludovicpages that's a problem with your logic. Nothing to do with fullpage.js.

ludovicpages:
@alvarotrigo, i never said it was a fullpage.js issue :)
Just trying to find a way to trigger the scroll when i want. 
Anyways, thanks for this library !

alvarotrigo:
@ludovicpages take a look at [my answer on stackoverflow](http://stackoverflow.com/a/36182162/1081396).

ludovicpages:
@alvarotrigo Awesome.
Actually, this is is the kind of logic i was working on afetr reading your previous comment.
There is a typo, you're setting animationIsFinish = true, but the var is animationIsFinished
Thanks a lot @alvarotrigo , it works great !

faliqaiman:
@alvarotrigo I tried using `onLeave` and yet still the delay is not working. Your answer at [stackoverflow](http://stackoverflow.com/questions/36176677/fullpage-js-adding-a-scroll-delay/36182162#36182162) helped but it just happen one time only.

And also I create a reproduction for my issue, and I've add fancybox plugins to see how it work. I've come out with [jsfiddle](https://jsfiddle.net/t3yzu3gs/) and also in [codepen](http://codepen.io/anon/pen/vGbqVG), but it show two different output. There's something worng with codepen, I guess, or Im doing it in a wrong way haha..

In jsfiddle, I structured by referring all your tips and trick from fullpage and stackoverflow. But in first section, I supposed when scroll down to next section, should wait the fancybox popup finished the effects but it just directly go to next slide while the effects is still running. Can you show the tricks for this one. Thanks.

alvarotrigo:
@faliqaiman you'd better ask for those questions in stackoverflow as they are not really issues but questions regarding specific usages of fullpage.js.

faliqaiman:
ok noted @alvarotrigo

PR #1213, opened 2015-05-12
Added forceDirection option which prevents backpeddling if desired (default false)

nsharmon:
I found myself in the circumstance in which the slides should flow one direction and prevent backpeddling, so I modified the existing library to support a new option forceDirection which does precisely that.  It won't let you move left (if there are slides) or up, and it won't let you move down unless all horizontal slides have been viewed.  It guarantees that the user will see every slide.  You're welcome to add this feature (or not add) as you wish.


alvarotrigo:
I believe you can achieve it by just adding this:

```javascript
$('#fullpage').fullpage({
    afterLoad: function (anchorLink, index) {
        setAllowScrolling(false, 'up');
    },
    afterSlideLoad: function (anchorLink, index, slideAnchor, slideIndex) {
        setAllowScrolling(false, 'left');
    }
});
```

nsharmon:
This would still allow you to move down without having viewed all horizontal slides.

alvarotrigo:
> This would still allow you to move down without having viewed all horizontal slides.

Right, that would require a bit more of code for the jQuery version of fullPage.js:

```javascript
$('#fullpage').fullpage({
    afterRender: function () {
        $.fn.fullpage.setAllowScrolling(false, 'up');
        $.fn.fullpage.setAllowScrolling(false, 'left');
    },
    afterLoad: function (anchorLink, index) {
                $.fn.fullpage.setAllowScrolling(isLastSlide($(this)), 'down');
    },
    afterSlideLoad: function (anchorLink, index, slideAnchor, slideIndex) {
        var section = ($(this).closest('.fp-section'));
        $.fn.fullpage.setAllowScrolling(isLastSlide(section), 'down');
    },
});

function isLastSlide(section) {
    if (section.find('.fp-slide').length) {
       return section.find('.fp-slide').last().hasClass('active'));
    }
    return true;
}
```

> Also, this solution doesn't yet seem to be possible for the pure javascript version.

Right. The javascript version is suppose to be a lighter and more simple version than the jQuery version. Adding the feature in the java-script version might not make much sense.

Issue #1044, opened 2015-02-24
Disable height when responsive is on 


borko-petkov:
Hi,
First i wan to say a big thanks for this wonderful plugin.
My problem is when responsive width is 1024 and when when the screen size is 1023 it turns page on normal scroll which is fine but its make every page 1023 height and i want to disable this. I want to make every section without defined height.
Is it possible?

alvarotrigo:
>  I want to make every section without defined height.

What do you mean with that.

borko-petkov:
When  screen size is 1023px i still have this extra style on each section style="height: 602px; background-color: rgb(240, 242, 244);". I want to remove this extra height when screen size is smaller than 1024.

alvarotrigo:
You can overwrite it.

```css
.fp-section,
.fp-slide,
.fp-tableCell{
    height: auto !important;
}
```

mezanddav:
I am sorry but this feature just doesn't make sense. The height: initial CSS is applied on the HTML and the body but the section keeps the height? In fact, it should be min-height...
I am having trouble going around this senseless feature.

alvarotrigo:
@mezanddav can you please elaborate your view on it?
There's a reason why sections keep the heigh value, in part, to be able to deal with `verticalCentered:true` and `scrollOverflow:true` options.

This feature was released way before the time flexbox was usable in most browsers, so perhaps we could review the implementation. 
But please, use constructive comments providing alternatives and solutions. Destructive comments won't help anyone. 

It would be nice if you could detail:

- What your issue is.
- Why you think the solution doesn't make sense.
- Provide alternative solutions.

Sorumeiji:
Hello @alvarotrigo,

I do have a issue with inline height being added when verticalCentered = { false }. Although it removes the fp-table classes. I still get added inline fixed heights to the sections. I don't think I can override this with !important,  Fixed heights cause things to overflow and break the responsive nature of the web. 

Using a min-height of 100%  for each section will ensure that each page is the full height, and it allows the browser to grow dynamically when content is added. I do understand there is the overflow and responsive height options, but it doesn't keep the same height always at 100% if you have different sizes of content on each page. 

I really love this plugin and it is saving by butt for a project, but I was wondering is there already an option to remove the fixed heights that are injected into the sections?

alvarotrigo:
> I really love this plugin and it is saving by butt for a project, but I was wondering is there already an option to remove the fixed heights that are injected into the sections?

[...]

Issue #919, opened 2014-12-17
Menu anchors highlight next section too early when using `height: auto !important`


noahstokes:
Hello. Excellent plugin, thank you for all your work on it. I've run into an issue where I am seeing the menu anchors highlight the next section too early. 

Here is my page: http://noahstokes.com/clients/bell/

*Steps to reproduce:*
If you scroll down to the 2014 By the Numbers you'll see the anchors jump to Excellence while you're about 2/3 of the way through the 2014 By the Numbers section. 

I've checked to make sure that none of the ID's match the anchors as the docs state, not sure what else I'm missing here.

Any advice is much appreciated!

alvarotrigo:
Just to point it out, this only happens when using:
```css
.section{
    height: auto !important;
}
```

TWSgit:
Is there any workaround, hack or fix for this issue already? :/ (I've run into the same effect and I have to fix it as soon as possible, but I have to use auto height sections either:/)

alvarotrigo:
> Is there any workaround, hack or fix for this issue already?

Nop, that's why it stills open. If you really need it to be fixed you can [ask for an urgent request](http://alvarotrigo.com/#contact-page).

TWSgit:
well I try to debug it myself at first (and contribute the fix if I can) but if I don't find a solution then I'll write.. thank you:) (I just thought perhaps there is a workaround already which isn't released yet or hacky (but works) which would be enough for me right now, but then I continue my debugging session;))

alvarotrigo:
Fixed in 2.7.7.

PR #1263, opened 2015-05-29
add a 'sectionAutoHeightSelector' option.

laomao800:
use for create some auto height section.
e.g. small footer

### DIFF
--- a/jquery.fullPage.js
+++ b/jquery.fullPage.js
@@ -40,6 +40,7 @@
 
     // section
     var SECTION_DEFAULT_SEL =   '.section';
+    var SECTION_AUTOHEIGHT =    'section-autoHeight';
     var SECTION =               'fp-section';
     var SECTION_SEL =           '.' + SECTION;
     var SECTION_ACTIVE_SEL =    SECTION_SEL + ACTIVE_SEL;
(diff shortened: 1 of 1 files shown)


laomao800:
<div class="section section-autoHeight" id="footer"></div>

alvarotrigo:
Thanks for it, I'll take a look at it!

luchorengo:
Great Work!! Thanks!!

alvarotrigo:
I believe a much simple approach would be to just use what I recommended [here](https://github.com/alvarotrigo/fullPage.js/issues/1044#issuecomment-75812112). 

It could be integrated in fullpage.js, but it would be a much more simple change in the code. 

In any case, if sections are not full height callbacks won't fire when expected as per [this open issue](https://github.com/alvarotrigo/fullPage.js/issues/919).

alvarotrigo:
> I believe a much simple approach would be to just use what I recommended here.

Sorry I just realized what I said doesn't make sense. The scrolling wouldn't work as expected. 

But still, I can not merge this without solving the related bug.

alvarotrigo:
I believe I will opt for this easier solution:

```
.fp-auto-height.fp-section,
.fp-auto-height .fp-slide,
.fp-auto-height .fp-tableCell{
	height: auto !important;
}
```

And then adding this single line in the javascript side:

```
var dtop = element.hasClass('section-autoHeight') ? (dest.top - windowsHeight + element.height()) : dest.top;
```

Seems easier and requires much less changes plus it works as well for horizontal slides, which your suggestion didn't.

alvarotrigo:
Implemented in 2.6.8.

PR #1273, opened 2015-06-02
Fix CDN link

mortonfox:


alvarotrigo:
Interesting. Seems like a recent change.
Thanks for it!!

Issue #1256, opened 2015-05-28
Responsive height


alvarotrigo:
Disabling `autoScrolling` when the viewport `height` is less than a value is something I find quite useful when dealing with small screens such as the ones of some tablets, which `width` can be quite big when turning the device into landscape position but their `height` is quite small. 

Something like the current `responsive` option working for the `width` could be done with the `height` as well.

salmanpatnee:
Could you please provide a sample code snippet it will be helpful.

alvarotrigo:
An option `responsiveHeight` has been introduced in fullPage.js 2.6.6.

PR #1274, opened 2015-06-02
- reponsive parameter can be a function which returns true or false

cemd:


alvarotrigo:
Explain yourself and why its necessary.

cemd:
In a project I needed to check not only the width but also the height of the viewport. Instead of adding another option I've added the ability to declare a function. The feature is also in another similar jQuery Plugin and I found it useful.

alvarotrigo:
The [responsive height](https://github.com/alvarotrigo/fullPage.js/issues/1256) option is a proposed enhancement. 

Checking the `height` inside the plugin initialization doesn't seem very reasonable. It wouldn't work on viewport resizing. 

It seems to me adding a function  `setResponsive(boolean)` can be more useful for that. 
This way you would be able to apply whenever you want (on mobile phone, under certain condition...)

cemd:
The height is not checked during initialization process, but everytime the plugin checks if it should go in responsive mode, the function is called (if provided). In there, I can finely adjust when I want the plugin to go responsive and when not. And it works on viewport resizing – I use it just like that in production. 

Check the responsive-Example, I've edited it to clarify the use of a function for this use case (however, I've only read the width as in the original one).

But: A `setResponsive` function seems also useful to me.

alvarotrigo:
> The height is not checked during initialization process, but everytime the plugin checks if it should go in responsive mode, the function is called (if provided)

Right. My mistake :D I don't know what was I thinking about.

The only problem I see with this method, although it seems useful, is that people with a basic knowledge of javascript might have troubles using it.

I see the `setResponsive` a more useful and versatile function to add.

alvarotrigo:
Now that I think about it, there's no need to create the `setResponsive` option, as basically is the same as `setAutoScrolling`. The only difference is that hides the vertical navigation bullets.

PR #1300, opened 2015-06-11
window.event can actually refer to an HTML element

djowett:
... in Firefox.  Code defensively against this.

This fix tested in Firefox & Chrome, but not IE.

### DIFF
--- a/jquery.fullPage.js
+++ b/jquery.fullPage.js
@@ -1038,7 +1038,7 @@
             //autoscrolling and not zooming?
             if(options.autoScrolling && !controlPressed){
                 // cross-browser wheel delta
-                e = window.event || e;
+                e = e || window.event;
                 var value = e.wheelDelta || -e.deltaY || -e.detail;
                 var delta = Math.max(-1, Math.min(1, value));
 
--- a/jquery.fullPage.js
+++ b/jquery.fullPage.js
@@ -1042,7 +1042,7 @@
                 var value = e.wheelDelta || -e.deltaY || -e.detail;
                 var delta = Math.max(-1, Math.min(1, value));
 
-                //Limiting the array to 150 (lets not waist memory!)
+                //Limiting the array to 150 (lets not waste memory!)
                 if(scrollings.length > 149){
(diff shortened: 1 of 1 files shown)


alvarotrigo:
What was the problem with it?

djowett:
I couldn't scroll down using the mousewheel

Using firebug I diagnosed that

window.event was refering to an <a id="event"  href="....">
rather than being undefined in MouseWheelHandler().


On 11/06/15 15:07, Álvaro wrote:
>
> What was the problem with it?
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/alvarotrigo/fullPage.js/pull/1300#issuecomment-111147030>.
>

-- 
Daniel Jowett

*Jowett Enterprises Ltd*
Tel: 0333 3 553 773
Mobile: 07870 667 126
daniel@jowettenterprises.com
www.jowettenterprises.com

alvarotrigo:
> I couldn't scroll down using the mousewheel

Mmm, which version of Firefox are you using and which OS ?
I'm testing the demo page in the latest Firefox (38.0.5) and it works as expected.

Also, I got not reports of this issue ever before.

alvarotrigo:
> window.event can actually refer to an HTML element

Wouldn't in that case fail anyway for IE ?

djowett:
FF 37.0.2 on Ubuntu. (Hence I can't easily test for IE)

I just tried that demo.
If I use Firebug to add    `id="event"`  to any html tag (e.g. `#infoMenu > ul`) it breaks.... i.e. scrolling no longer works.

alvarotrigo:
Interesting... I didn't know you could get the elements of the page in that way in Firefox.

Still, I wouldn't recommend you to do it.
Luckily IE doesn't get the DOM element that way and the plugin won't crash.

I'll merge it for this very specific cases. Thanks for it!

djowett:
I didn't know that about Firefox either until today & it seems not very good of Firefox IMO, but it's good to avoid the issue where we can.

Thanks!

Issue #1337, opened 2015-06-28
lazy-loading per slide


blueyed:
The current lazy-loading mechanism loads all elements for a section, but it would be better to do this per slide.

In https://github.com/alvarotrigo/fullPage.js/issues/1289#issuecomment-109956072 you mention a planned `lazyDistance` feature, which is probably meant to pre-load elements?

In my case I have 30 images (in 30 slides), and it would be great if only the current and prev/next images would be loaded.

blueyed:
#1338 got merged.
I've added #1340 as a follow-up with a new `lazyDistance` option (to be discussed).

PR #1338, opened 2015-06-28
lazyLoad: per slide

blueyed:
This is a first shot at addressing https://github.com/alvarotrigo/fullPage.js/issues/1337.

### DIFF
--- a/jquery.fullPage.js
+++ b/jquery.fullPage.js
@@ -1342,6 +1342,10 @@
         */
         function lazyLoad(destiny){
             //Lazy loading images, videos and audios
+            var slide = $('.fp-slide.active', destiny);
+            if( slide ) {
+                destiny = $(slide);
+            }
             destiny.find('img[data-src], video[data-src], audio[data-src]').each(function(){
                 $(this).attr('src', $(this).data('src'));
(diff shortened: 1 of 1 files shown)


alvarotrigo:
Thanks for it!
Just a couple of comments below.

blueyed:
Thanks for your prompt response!  I'm glad you like it and that the approach seems to be good.

I've pushed a "fixup!" comment.

Please let me know if it's good like that, and I'll squash it.

I've only quickly looked into this when I've realized that it needs to be optimized, and would have expected that my attempt wouldn't be that good probably.

blueyed:
What do you think about the idea from #1337 to pre-fetch the previous/next slide(s)?

This should be configurable as a number for the amount of surrounding items to pre-fetch, and default to `1` for next/prev, allowing for `0` to only load the current slide then.

Is this what `lazyDistance` was meant to be?
It could then apply to sections and slides maybe.

alvarotrigo:
Thanks for it, merged.

blueyed:
Thanks for merging it, although it was not really ready.

See #1340 for a follow-up regarding the `lazyDistance` idea.

PR #1340, opened 2015-06-29
Lazy slides: lazyDistance setting

blueyed:
Followup to https://github.com/alvarotrigo/fullPage.js/pull/1338, adding a `lazyDistance` option.

Ref: https://github.com/alvarotrigo/fullPage.js/pull/1338#issuecomment-116347494


syd2011:
coool

blueyed:
Rebased on master.

@alvarotrigo 
What do you think about it?

alvarotrigo:
Looking good!! 

Although that's lazy distance for slides.
Ideally lazyDistance will be for both, slides and sections. 

Also, consider wrapping all of it in a function. Something like:

```
destiny = getDestinationsToLazyLoad(destiny);
```

Oh, and, the condition `if( slide.length ) {` seems to be there to only deal with the first slide of each section when landing in any slide coming from other vertical section.

Your code should also be applied when moving from slide to slide. So I believe it shouldn't be placed inside that condition.

PR #1343, opened 2015-06-30
Pass down container to call of `afterRender`

blueyed:
`this` is the `window` object currently, although the documentation
states that it should be the plugin container.


alvarotrigo:
You don't need to pass it. `container` is global.

blueyed:
I see.  I've amended the commit.

alvarotrigo:
Thanks for it!

Issue #950, opened 2015-01-08
browser back-button / history problem with first section due to missing anchor-hash.


tripletto:
First of all: great work. Thank you very much.

After fiddling around a bit with fullpage.js, I encountered a little problem which can be reproduced with your demo. 

Assume we have 3 or more sections. When I scroll from the first section (0) to the second section (1) and I then decide to go back to the first section via browser back button, the browser back-button won't work. The first section is missing its anchor / hashtag.

I fiddled a bit around and thought I could use your afterLoad-callback to add the missing anchor like this, so the browser navigation buttons will work from the first section on: 

```javascript
afterLoad: function(anchorLink, index){
            //using index
            if(index == 1){
               location.hash = 'section0';
            }};
 ```

The problem I ran into was, afterLoad seems to work only from 2nd section on upwards. The principle was tried out with afterRender and works but isn't very usefull because of deeplinking and stuff...

So i guess I would need something along the lines of

```javascript
afterRender: (anchorLink, index) {
 //stuff
};
```

So one could check if the first section is active and if so give it the missing hashtag. Does that make sense to you or am I totally stupid? :D Is there a method to do this that I have overlooked?

Best regards
tripletto

alvarotrigo:
I believe the only way to solve it properly would be to force the addition of the anchor of the default active section on page load. 
Which I'm not quite sure people would like.

Canubiz:
I also just ran into this problem mentioned by the thread opener. I am not 100% sure but think I fixed it by fiddling around in the jquery.fullpage.js file. Hopefully it doesn't break any other functionality - but so far it works… Would be interested to hear what you think about this. :)

This is my hastChangeHandler function:
```javascript
function hashChangeHandler(){
            if(!isScrolling){
	            
                var value =  window.location.hash.replace('#', '').split('/');
                //alert('hash is ' + value);
                var section = value[0];
                var slide = value[1];

                if(section.length){
                    //when moving to a slide in the first section for the first time (first time to add an anchor to the URL)
                    var isFirstSlideMove =  (typeof lastScrolledDestiny === 'undefined');
                    var isFirstScrollMove = (typeof lastScrolledDestiny === 'undefined' && typeof slide === 'undefined' && !slideMoving);

                    /*in order to call scrollpage() only once for each destination at a time
                    It is called twice for each scroll otherwise, as in case of using anchorlinks `hashChange`
                    event is fired on every scroll too.*/
[...]

PR #1365, opened 2015-07-14
"location.hash = the-hash" not working in Android 4.0.3

HeGanjie:
This glitch will cause the URL remains unchanged after the user slid the page, and back button not working as usual. But strangely it works on Android 2.3.3
This answer seems related: http://stackoverflow.com/a/6975488/1745885

### DIFF
--- a/jquery.fullPage.js
+++ b/jquery.fullPage.js
@@ -2078,7 +2078,7 @@
         */
         function setUrlHash(url){
             if(options.recordHistory){
-                location.hash = url;
+                location.href = '#' + url;
             }else{
                 //Mobile Chrome doesn't work the normal way, so... lets use HTML5 for phones :)
                 if(isTouchDevice || isTouch){


alvarotrigo:
Have you tested it in other devices (Windows Phone, iPhone... ) and browsers? (IE, Safari, Chrome..)

The link you provide in stackoverflow talks about *window.location.hash = location.hash” not working on Webkit browsers included Safari. And that is not true. It works as expected on safari using iPhone. 

That makes me question the validity of the provided answer as this statement doesn't make any sense to me:

> Webkit has two oddities that prevent window.location.hash = location.hash from working normally

alvarotrigo:
The [answer in stackoverflow](http://stackoverflow.com/a/6975488/1081396) is from 2011 and the [reported bug](http://www.quirksmode.org/bugreports/archives/2005/06/Setting_locationhash_to_the_same_value_twice_cause.html) from 2005.

alvarotrigo:
I have also tested the [demo page](http://alvarotrigo.com/fullPage/)  in www.browserstack.com in Samsung Galaxy S6 and Chrome and it seems to work as expected as well.

Why device, browser and version of Android were you using to reproduce the problem?

HeGanjie:
I've tested only under Android emulator 4.0.3 & 2.3.3, Chrome device mode and iOS 8.4 (Safari & Chrome). Result:

|  	| Android emulator 4.0.3 	| Others 	|
|-----------------------	|---------------------------------------------------	|--------	|
| location.hash=url 	| demo page sometimes ok, not working in my project 	| ok 	|
| location.href='#'+url 	| ok 	| ok 	|

alvarotrigo:
> demo page sometimes ok,

Sometimes? What do you mean. 

Have you tried it directly in the real device? Emulators are not perfect sometimes.

HeGanjie:
I finally find out this isn't the point of my problem. 

My problem is: The first time user scrolled to another page, he can not return to the first page by clicking the back button. So I append a line to workaround `location.href = '#slide';`. So here is the point, Android  4.0.3 emulator wouldn't "pushState" after that line executed, so the user still can't back to the first page.

```javascript
$('#fullpage').fullpage({
  paddingBottom: '3.1rem',
  anchors: ['slide'],
  controlArrows: false,
  verticalCentered: false,
  loopHorizontal: false,
});
$.fn.fullpage.setAllowScrolling(false);
location.href = '#slide';
```

But I still have no idea why it works after I change `location.hash=url` to `location.href='#'+url`

alvarotrigo:
Your issue seems to be exactly the same as [this open issue](https://github.com/alvarotrigo/fullPage.js/issues/950).

PR #1369, opened 2015-07-15
IE9 has both mousewheel and wheel event

haledeng:
In IE9, both mouse wheel and wheel will be listened, as a result, the event handler will be done twice.


alvarotrigo:
I have no idea what is all this mess :D I'd better close it...

PR #1391, opened 2015-07-23
added option `fitToSectionDelay`

Suven:
Hey,

when using fitToSection fullPage.js currently uses a hardcoded 1000ms-delay.
This PR introduces a simple option for said delay.

The changes are not yet minified as I was not sure what exactly you are using for that :/


alvarotrigo:
Thanks for it!

I'll take a look at it in a few days for the next release.

PR #1396, opened 2015-07-26
A little trouble when using fullPage.js with AMD (Require.js)

spalax:
When i use fullpage.js with require.js window.load event triggers before fullpage has registered handler for scrollToAnchor. So i resolved this issue by checking for document.readyState, and it is helps.

### DIFF
--- a/jquery.fullPage.js
+++ b/jquery.fullPage.js
@@ -659,10 +659,11 @@
             //setting the class for the body element
             setBodyClass();
 
-            $window.on('load', function() {
+            if (document.readyState != "complete") {
+                $window.on('load', scrollToAnchor);
+            } else {
                 scrollToAnchor();
-            });
-
+            }
         });
 
 


alvarotrigo:
I'll take a look at it in a few days.

alvarotrigo:
It seems to me that this will fire the scrollToAnchor function when the site is not yet loaded:

```javascript
if (document.readyState != "complete") {
    $window.on('load', scrollToAnchor);
}
```

alvarotrigo:
@spalax anything to comment on my last answer ?

spalax:
@alvarotrigo sorry, it is a bit strange, because, if you will load your script synchronously at the html body i guess this condition will be (document.readyState != "complete") true alway. But if you require(fullpage.js) asynchronously it is mean that sometimes, fullpage.js might be required after LOAD event fired.

alvarotrigo:
The solution doesn't seem to be simple.
Check this answer: http://stackoverflow.com/q/31308059/1081396

alvarotrigo:
I would argue that it doesn't make much sense to load fullpage.js dynamically as it is necessary for the correct visualization of the page in any case.

spalax:
You might controll loading manually, so if you show loading image which is playing until all async libs are loaded. In my opinion it is not a good practice to control such a thing inside library exclusively.  
IMO scrollToAnchor should be possible to run manually and of course i guess it is a trouble if your library execution logic depends on the loading order =(((

spalax:
IMO It is a bit brakes Separation of Concerns ...

alvarotrigo:
I believe this got solved by this PR: https://github.com/alvarotrigo/fullPage.js/pull/1982

Issue #1403, opened 2015-07-28
Destroy() doesn't clear html/body classes and inline styles


SergeRNR:
After destroy() method call helper classes and inline styles are not cleared:
html : class= fp-enabled, style=overflow: visible; height: initial;
body: class= fp-viewing-0, style=overflow: visible; height: initial;

Demo from DOCs, please inspect iFrame:
http://codepen.io/alvarotrigo/pen/bdxBzv

alvarotrigo:
Thanks for reporting it.
I'll take a look at it for the next release.

ziggythehamster:
#1414 fixes the style, #1418 fixes the classes

alvarotrigo:
Solved.

PR #1414, opened 2015-08-04
[Patch] [Test] Clean up html and body tags on destroy

ziggythehamster:
This PR makes fullPage.js clean up the `<html/>` and `<body/>` inline styles and adds a test that ensures this is happening (run it on the old version and see that it fails). The test could probably also check the container, but I'm unsure what that is in the test file (I think `$('#fullpage')`)

Your package.json doesn't include a build script, so I did not update the .min.js file.

### DIFF
--- a/jquery.fullPage.js
+++ b/jquery.fullPage.js
@@ -2393,6 +2393,11 @@
                 'touch-action': ''
             });
 
+            $htmlBody.css({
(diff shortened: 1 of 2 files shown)


alvarotrigo:
Thanks for it!
Related with #1403

PR #1418, opened 2015-08-04
[Patch] [Test] Clean up more things

ziggythehamster:
Noticed that fullPage.js was not cleaning up the classes on `<html/>` and `<body/>`, so added a test and then fixed the bug.

I *think* I've got fullPage.js cleaning everything up now. :)


alvarotrigo:
Thanks for it!
I'll merge it in in the next release together with other improvements.

Issue #1426, opened 2015-08-08
Navigation not appearing


jacoblindgren:
Hello,

I just began using fullPage.js and I hadn't been able to get the navigation to appear, despite setting navigation: true, and navigationPosition: 'right'. I asked a [question](http://stackoverflow.com/questions/31887893/certain-options-not-taking-effect-in-fullpage-js/31888127) on Stack Overflow as to why this was happening and someone replied with a solution that ended up working in addition to why the problem exists. (Something about setting the 'nav' selector *before* it adds the element to the body).

It's a little over my head, but I figured I would post here in case it turned out to be a proper bug. If not, ignore me. 

Thanks.

alvarotrigo:
Right. Thanks for reporting it! I'll solve it soon.
It seems to be a bug introduced in the latest version 2.6.7.1 after refactoring. 

You can use 2.6.7 instead, which doesn't have the bug, and has exactly the same features as 2.6.7.1.

In fact, the minified version was not updated and you can still use the master branch with it.

alvarotrigo:
Solved.

PR #1427, opened 2015-08-08
FIXED: #1426 - Navigation not appearing

maheshwaghmare:


alvarotrigo:
Just inverting the order of the first two rows would do it:

```
$body.append('<div id="' + SECTION_NAV + '"><ul></ul></div>');
var nav = $(SECTION_NAV_SEL);		
```

This way the selector would be cached.

alvarotrigo:
Solved in 2.6.8

PR #1450, opened 2015-08-16
Add Background Image

SamStone92:
Added the ability to add a background image for each section


alvarotrigo:
Thanks for it, although I don't think it will be merged. 
That should be done in CSS. There's no need to add it in the plugin. 

It would also have to be done for the slides, which make it more complicated yet. CSS its its place.

PR #1467, opened 2015-08-25
Allow Custom Slides Nav Element

josiahsprague:
This allows someone to create custom slides nav markup by manually inserting an element with the class .fp-slidesNav in the section by looking for an existing .fp-slidesNav before appending the default.

### DIFF
--- a/README.md
+++ b/README.md
@@ -321,7 +321,7 @@ $('#fullpage').fullpage({
 
 - `showActiveTooltip`: (default `false`) Shows a persistent tooltip for the actively viewed section in the vertical navigation.
 
-- `slidesNavigation`: (default `false`) If set to `true` it will show a navigation bar made up of small circles for each landscape slider on the site.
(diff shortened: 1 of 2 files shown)


alvarotrigo:
Thanks for it.
But it doesn't sound very clear or straight forward to me...

Yeah, they can create their own markup but... they still need to use some compulsory elements to fit in fullPage.js, such as they have to contain a link `<a href`, they have to be in a list `li`...
Otherwise the links to each slide won't work at all.

I don't see it easier than overriding the default css styles applied by the plugin. That's what I could see from the code. Correct me if I'm wrong.

josiahsprague:
In my case, I need to add numbers inside each element, add a custom class, and remove the `<span>`. This is what my markup looks like;
```
<div class="controls__nav fp-slidesNav bottom">
  <ul>
    <li><a href="#" class="active">1</a></li>
    <li><a href="#">2</a></li>
  </ul>
</div>
```

Issue #1442, opened 2015-08-13
Vertical window resize bug


podorozhny:
Hello.
Vertical window resizing causes a bug with slides: active slide gets out beyond the browser window. Scrolling to next slide fixes the problem.
Tested in chrome and firefox last versions (ubuntu 14.04) on http://alvarotrigo.com/fullPage/

alvarotrigo:
Introduced in the last release 2.6.9 due to the fix of Chrome bug.

voxes:
I am a JS novice, so forgive me if this is obvious or non-relevant: Could this issue be partially caused by calling promise().done() on an animation in performMovment() that would be scheduled but not completed due to performMovment() being inside the requestAnimationFrame() callback function, causing the function to hang until a scroll event forced the animation repaint and thus satisfied the promise().done()? When I tested the function inside the promise().done() with a simple console line, it was not running after a re-size until I fired a scroll event.

alvarotrigo:
@voxes if you think you have a solution please commit a pull request.

Jimmy-Codes:
Hey there, I'm having this issue as well. It becomes very apparent when opening up Chrome's dev tools and will surely show itself to our users when they attempt to adjust the window to experience the full page. Is there an older version that is stable and does not have this vertical resize issue?

alvarotrigo:
> Is there an older version that is stable and does not have this vertical resize issue?

As I said, it was introduced when fixing another major problem with Chrome users.

Jimmy-Codes:
Wishful thinking. Thank you for your plugin and help.

alvarotrigo:
@Jimmy-Codes Hopefully Chrome will fix its bug soon and we can get rid of this secondary one.

alvarotrigo:
@Jimmy-Codes just to let you know, you can also fix the problem by using 2.6.8 and, rather use `css:false` or what I would recommend: `scrollBar:true`.

Jimmy-Codes:
Wow, that looks like it worked great! How do I know if it works cross-browser-version, though? I'm on Chrome 44.0.2403.157 (64), and don't seem to get any issue.

Also, is `scrollBar: true` supposed to show the page as a normal scroll? I'm still seeing my intended "slide" view above the mobile threshold, which is perfect, but I wanted to make sure this is all correct given the option's name.

Thank you!

alvarotrigo:
> How do I know if it works cross-browser-version, though?

[...]

Issue #1389, opened 2015-07-22
Chrome 44.0 issue


gwixted:
Have been building a site using fullPage and everything was working great until today (Chrome 44.0 was released yesterday). Now, with autoScrolling enabled (the default behaviour) it appears to jump to the first section each time i scroll (using the mouse) and scrolls through all sections until it reaches the 'next' section. It's even happening on the fullPage demo page.

alvarotrigo:
Can't reproduce it in [the demo page](http://alvarotrigo.com/fullPage/). 
Please provide a reproduction.  Closed until then.

alvarotrigo:
Thanks for reporting it. I'll investigate it now.

alvarotrigo:
Have you tried it on Mac or Windows?
I saw the problem taking place in Mac. Not sure about windows yet.

gwixted:
I'm only seeing it on Mac. Just tested it on Windows (Windows 7, Chrome 44.0) and it appears to be working just fine.

alvarotrigo:
Mmmm weird... are you using mouse or trackpad ?

gwixted:
Mouse on Mac, trackpad on Windows.

alvarotrigo:
Ok thanks. I keep investigating it.

alvarotrigo:
Seems to be a tough one... It seems to me that there's some kind of bug in Chrome with translat3d, but I can not isolate it and reproduce it.

alvarotrigo:
The problem doesn't take place when using keyboard to navigate up or down. 
I've reduced the plugin to the basics (400 lines) and still can not detect where the problem is...

alvarotrigo:
Here I have fullpage.js reduced to 100 lines and being able to reproduce the bug by just scrolling down.
http://jsfiddle.net/6fdLk7fo/

As you can see, the first section will become visible again when scrolling from the 2nd to the 3rd section. 
This shouldn't take place.

alvarotrigo:
A much cleaner reproduction of the issue:
http://jsfiddle.net/9ksx000q/3/

alvarotrigo:
It happens in Windows as well :(

mcadam:
+1 same problem using Mac and trackpad

alvarotrigo:
I opened a question in stackoverflow hoping somebody else can realise about it or propose any solution:
http://stackoverflow.com/questions/31575177/bug-with-translate3d-in-chrome-v-44

[...]

PR #1469, opened 2015-08-26
Fixed the vertical resizing issue caused by async.

voxes:
This is a fix for issue #1442 

Moved the local variable cache assignment out of `requestAnimationFrame` to avoid them being overwritten before being saved.

Also changed `!isResizing` to `!v.localIsResizing` in first check inside `requestAnimationFrame`.

Re-sizing bug no longer occurs, although the slides do animate after re-size instead of instantly snapping into position. I'm not sure if this was the original intended effect or not, so I'm leaving it as-is. The `css3: false` was for testing, it works with both CSS3 and JQuery animations.

### DIFF
--- a/examples/simple.html
+++ b/examples/simple.html
@@ -21,7 +21,8 @@
 	<script type="text/javascript">
 		$(document).ready(function() {
 			$('#fullpage').fullpage({
-				sectionsColor: ['#f2f2f2', '#4BBFC3', '#7BAABE', 'whitesmoke', '#ccddff']
+				sectionsColor: ['#f2f2f2', '#4BBFC3', '#7BAABE', 'whitesmoke', '#ccddff'],
+				css3: false
 			});
 		});
 	</script>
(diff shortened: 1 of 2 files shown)


alvarotrigo:
I believe this will produce [the popular bug in Chrome 44](https://github.com/alvarotrigo/fullPage.js/issues/1389).

voxes:
Will test for it and make adjustments if possible.

alvarotrigo:
I would probably wait until the next version of Chrome is released before taking any deeper look at it. 
If the next Chrome version fixed its bug, then we won't have to worry about this bug either as it will automatically dissappear when removing `requestAnimationFrame`

eolant:
It shouldn't conflict with Chrome bug, because we just creating object and doesn't assign any translate3d anywhere.
Also if you want to remove animation you should change this in performMovement function:
```javascript
transformContainer(translate3d, v.localIsResizing ? false : true);
```
After that fix, +1

alvarotrigo:
The issue was solved in 2.7.1.

PR #1500, opened 2015-09-11
Force string for anchor names

johanhermansson:
I used numbers in `data-anchor` and got an error. The `text` variable was an integer which does not have the `replace` method. I fixed it with forcing the variable to be a string.

You want it in another branch, or in another way? :)

### DIFF
--- a/jquery.fullPage.js
+++ b/jquery.fullPage.js
@@ -2199,6 +2199,9 @@
             if(slide.length){
                 text = text + '-' + slideAnchor;
             }
+            
+            //force string
+            text = '' + text;
 
             //changing slash for dash to make it a valid CSS style
             text = text.replace('/', '-').replace('#','');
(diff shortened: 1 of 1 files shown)


alvarotrigo:
I would use the same approach used a couple of lines above:

```javsacript
var text = String(sectionIndex);
```

Just cast it here as well:
```javascript
if(slide.length){
    text = String(text + '-' + slideAnchor);
}
```

Same branch is ok.

alvarotrigo:
Well, in fact I would just move the existing one under the condition `if(slide.length)`:

```javascript
if(slide.length){
    text = String(text + '-' + slideAnchor);
}
var text = String(text);
```

johanhermansson:
Ah, yes! That looks nicer! Pull request updated :)

alvarotrigo:
Thanks for it!

PR #1514, opened 2015-09-16
Scroll slides without animation after resize

eolant:
Since it's scrolling sections without animation I think it should do the same with slides.


alvarotrigo:
I'll take a look at it for the next release.
Thanks!

PR #1498, opened 2015-09-09
Moves scrollOverflow code to separate handler object

rmarscher:
This update adds a new option called 'scrollOverflowHandler'
which consolidates all code related to scrollOverflow and
allows it to be swapped out with different functionality.

By default, it will use the existing jquery.slimScroll handling
of scroll overflow.

This should make it easier for users to either customize the
slimScroll handling or replace it with other code just as wrapping
content in a div with `overflow: scroll` or utilizing iscroll.js
as requested in issues #506, #612 and others.

PR #612, opened 2014-08-03
Added slimScroll default and extending options

Kwaadpepper:
Slimscroll options to make fullpage more usable

Issue #506, opened 2014-06-13
scrollOverflow doesn't work well on touch devices


vmahabir:
Hi Alvaro, 

First of all; nice job on Fullpage!
I'm implementing it in a responsive Boilerplate template. It seems to work, however I encounter some problems during testing. 

When scrolling (autoscroll = true) from the first slide to the second and to the third; works perfect The fourth and last slide contains the full website content so the fourth slide is bigger than the other slides. I've implemented it in the last slide so there has to be a scrollable area because it's larger than the other three slides. The scrolling in the scrollable area seems very choppy. autoscrolling and scrollOverflow is true. Do you have an idea what's causing this choppy scrolling?  On a mobile device it's really bad..

Also, when scrolling from the third slide to the fourth slide: it looks like the third slide is overlapping the fourth slide at the top while scrolling down to the fourth slide.

Do you have an idea about what could happends at that moment? And can I fix it in some way?
You could view a demopage here: http://vishumahabir.com/test/tmc/fpjstwitter/html/

Thank you!

webdrifter:
This helped me:
http://stackoverflow.com/questions/26334827/slimscroll-bar-very-slow-in-mobile-browser
Downloading v1.3.7 in https://github.com/rochal/jQuery-slimScroll/releases

alvarotrigo:
With the change of slimScroll to iScroll.js in 2.8.0 this issue has been solved.

PR #1543, opened 2015-09-25
Moves scrollOverflow code to separate handler object

rmarscher:
This update adds a new option called 'scrollOverflowHandler'
which consolidates all code related to scrollOverflow and
allows it to be swapped out with different functionality.

By default, it will use the existing jquery.slimScroll handling
of scroll overflow.

This should make it easier for users to either customize the
slimScroll handling or replace it with other code just as wrapping
content in a div with overflow: scroll or utilizing iscroll.js
as requested in issues #506, #612 and others.

This pull request is a re-opening of #1498 against the new "dev" branch.

### DIFF
--- a/jquery.fullPage.js
+++ b/jquery.fullPage.js
@@ -85,6 +85,8 @@
     var $window = $(window);
     var $document = $(document);
 
+    var defaultScrollHandler;
+
     $.fn.fullpage = function(options) {
 
         // common jQuery objects
@@ -120,6 +122,7 @@
             continuousVertical: false,
             normalScrollElements: null,
             scrollOverflow: false,
+            scrollOverflowHandler: defaultScrollHandler,
             touchSensitivity: 5,
             normalScrollElementTouchThreshold: 5,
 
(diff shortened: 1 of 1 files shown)


alvarotrigo:
Thanks!

rmarscher:
Awesome. Thank you!!

anywayalive:
Great update! How soon should we see it in the master branch so it can be updated with bower?

alvarotrigo:
At least a week. 
I would like to test it with iScroll.js and probably add the option of using it together with this update.

rmarscher:
Here's that gist again for an iScroll adapter - https://gist.github.com/rmarscher/5f3a62b960f5f903a6af
That worked for my use case. But I didn't try it with horizontal slides.

alvarotrigo:
Thanks for it @rmarscher  I'll take a look whenever I find some time for it.

alvarotrigo:
@rmarscher don't you need `mouseWheel: true` for your [iScroll gist](https://gist.github.com/rmarscher/5f3a62b960f5f903a6af#file-iscrollhandler-js-L50)? 

Check out [this demo](http://jsfiddle.net/Lehqx9ne/5/) I made. It won't scroll unless you turn it on.

rmarscher:
@alvarotrigo If I recall from the project that used this, I had `overflow-y: scroll; -webkit-overflow-scrolling: touch;` on the #wrapper here and only used iScroll for touch handling.

alvarotrigo:
@rmarscher  Yeah, that will show a default scroll bar. Which looks ok for Mac, but not for Widows. 
In any case, it might be nice to provide iScroll as an option in case someone is interested I guess.

ASchimek:
Are there any news? Or should I try to implement iscroll as described in the gist above?

alvarotrigo:
iScroll project doesn't seem very active anymore either. I got no response on [the topic I opened](https://github.com/cubiq/iscroll/issues/980) and it seems the only way of adding it to fullpage.js would be to use a normal scroll bar for windows devices.

You can use the GIST that @rmarscher provided and make use of the `dev` version of fullpage.js where the pull request was merged.

PR #1559, opened 2015-09-30
Add prefix for older browser versions

TheLand:
Adding prefixes to requestAnimationFrame extend support to older browser.
In my case i tested in FF20.


alvarotrigo:
Thanks for it!
Just out of curiosity. Do people still using Firefox 20?? We are in version 41!

TheLand:
I can't also explain it myself. But Banks seems to still use it! Thanks to you for this great plugin!

PR #1560, opened 2015-09-30
Add minDelta value and exclude node_modules

Ahrengot:
Hey Álvaro,

## Delta and scrolling on Mac OS X
Thanks for creating this awesome script! I noticed that scroll will often trigger unexpectedly on Mac OS X so I added a minDelta value to work around that. It’s set to a default of 20 which seems like a nice balance between not having to scroll too hard and not accidentally causing unwanted scrolling.

## Node modules
I noticed that node_modules weren't excluded from the repo so I went ahead and did that.

## Minification
I couldn't find any minification task so the didn't compress the code. *That probably means you shouldn't accept this pull request just yet*, but if you tell me how you prefer the code to be compressed (Which tool to use and any options I should enable)  then I'll send a new pull request your way.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 .DS_Store
 bower_components/
+node_modules
--- a/jquery.fullPage.js
+++ b/jquery.fullPage.js
@@ -1075,7 +1075,7 @@
                 var isScrollingVertically = (Math.abs(e.wheelDeltaX) < Math.abs(e.wheelDelta)) || (Math.abs(e.deltaX ) < Math.abs(e.deltaY) || !horizontalDetection);
 
                 //Make sure user intended to scroll
-                if (Math.abs(value) < options.minDelta) {
+                if (options.lockAnchors && Math.abs(value) < options.minDelta) {
                     return;
                 }
 
--- a/jquery.fullPage.js
+++ b/jquery.fullPage.js
@@ -122,6 +122,7 @@
             scrollOverflow: false,
             touchSensitivity: 5,
             normalScrollElementTouchThreshold: 5,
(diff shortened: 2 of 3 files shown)


alvarotrigo:
Thanks for it @Ahrengot, although the minDelta option doesn't seem like a proper solution.
The value of delta is different for different browsers, if you try it in Firefox you'll see how you get much lower values (3, 6...)

You can check it here:
http://jsfiddle.net/oo8ngfpL/2/

alvarotrigo:
Check this out: 
http://stackoverflow.com/questions/5527601/normalizing-mousewheel-speed-across-browsers

Ahrengot:
I might be missing something here, but I get the same delta in safari, chrome and firefox. I also get really low values in Chrome/Safari like 3.

I'm using a magic trackpad and a magic mouse and get similar responses on both.

We could do delta normalization if you're up for it. I wouldn't mind spending some time looking into that.

alvarotrigo:
You have to test it with a non kinetic scrolling device. A normal mouse with a normal mousewheel.

Ahrengot:
Ah, I misunderstood your last comment. I thought you meant that different browsers on OS X send back different deltas. 

Anyway, are you opposed to a minDelta option in general, or do you want me to explore normalizing the deltas across operating systems and hardware inputs?

alvarotrigo:
> Anyway, are you opposed to a minDelta option in general, or do you want me to explore normalizing the deltas across operating systems and hardware inputs?

I'm open to improvements.
If you mange to normalize it and improve the plugin, I'll be open to that. If its possible or not, I'm not quite sure yet.

Issue #1558, opened 2015-09-30
Build script for compilation and compressing of SCSS, CSS and JS files


d13:
Providing a build process will ensure parity in the compilation and compressing of files.

I have a working version of this on my fork: https://github.com/d13/fullPage.js

This also addresses #1096

Thanks and keep up the great work!

alvarotrigo:
Sounds good to me!
I'm not familiarized with gulp, but feel free to commit the pull request.

alvarotrigo:
Merged here: https://github.com/alvarotrigo/fullPage.js/pull/1570

Issue #1096, opened 2015-03-20
minified css please?


tomByrer:
I was going to automate uploading of your lib to (jsDelivr CDN](http://www.jsdelivr.com/#!jquery.fullpage), but noted you only minify the JS.  Could you also minify the CSS so I can program the bot to auto-update?

alvarotrigo:
I usually try to keep it as simple as possible for newbies and adding more files can make it a bit more confusing for this group.
I don't consider a minification of the existent CSS file will provide much benefits, so I'm not quite sure about this idea.
Also, other CDNs such as cdnjs.com get all files and compress them when needed:
https://cdnjs.com/libraries/fullpage.js, which sounds like a good alternative solution to me.

tomByrer:
Many (majority?) of Github repos use a `/dist` folder or the like for the minified files.
Minifying your CSS will save 2.5k, adds up.
I thought about programming the bot to minify CSS, but I'm apprehensive about breaking the CSS; I'd hope the dev would test minified files before publishing.

[jquery.reveal](http://www.jsdelivr.com/#!jquery.reveal) or [reveal.js](http://www.jsdelivr.com/#!reveal.js)?  impress is there also.  Humm... 2/3 of those are providing non-minified files, but I think they were uploaded before minification enforcement.

BTW, I prefer minified files include a comment-header like you do for your `.min.js`, since we can [concoct files](http://cdn.jsdelivr.net/g/jquery@1.11,jquery.ui@1.9.2%28jquery.ui.effect.min.js%29,slimscroll@1.3,jquery.fullpage@2.4), saving several HTTP requests & a bit smaller compression.

alvarotrigo:
> jquery.reveal or reveal.js? impress is there also.

Nop, I'm talking about [reveal.js](https://github.com/hakimel/reveal.js) or [impress.js](https://github.com/bartaz/impress.js/), which are different from the files you can find in jsdelivr.


> I thought about programming the bot to minify CSS, but I'm apprehensive about breaking the CSS

How would you break it? Not quite sure to understand it. A proper compression should provide the exact same results as the non minified version. In the case of using CSS, by just removing whitespaces, indentation and line jumps.

tomByrer:
> I'm talking about reveal.js ... , which are different from the files you can find in jsdelivr.

Uploaded by a human being who compressed said files on their own, which I don't want to do any more, esp for someone like you who publishes new versions almost weekly (which is a good thing).
[...]

PR #1570, opened 2015-10-04
Build tasks for compiling SCSS & compressing CSS and JS files

d13:
A high-level description of changes made:
- added Gulp and Gulp plugins
- updated the SCSS styles to match the styles in the CSS file
- added the `dist` directory to unobtrusively write the minified files and source-maps

Addresses: #1558 & #1096


alvarotrigo:
Thanks for it @d13! Looks good!
Although, not sure why you made that many changes in the CSS file. It seems like you changed the order. Why? Oh, and the indentation as well. 

I saw you also changed the absolute links to relative ones in the documentation such as this one:
````
(https://github.com/alvarotrigo/fullPage.js#introduction)
```
Could you revert those changes as well? Keeping them absolute helps in case some sites copy the readme content into theirs or even for sites such as npmjs which extracts the docs from github. I prefer them to always link to the github docs.

If you revert those last changes and the CSS ones I'll merge it.  Feel free to create a new pull request for clarify.

alvarotrigo:
fullpage.js 2.7.4 has just been released. You'll probably have to update your copy now before commiting it again.

d13:
Merged and reverted style changes.

The only CSS change I kept was the change of the copyright comment from `/**` to `/*!` so the compression tool will preserve that comment.

alvarotrigo:
Thanks a lot!

alvarotrigo:
@d13 I'm not used to work with sass or scss but there's something wrong on the gulp file.

You said in the read me

>// Only compile the SCSS
>gulp scss

But `gulp sass` is what its required instead.
Not quite sure what are you doing there though.

d13:
The mistake is in the documentation, I should have written `sass` as the task name as that is technically what is being compiled. `.scss` is just the file extension for SASS.

d13:
I'll push a fix for this shortly.

alvarotrigo:
@d13 what are you really doing with the SCSS file? Compiling it to `.css`?

In that case we should better get rid of it, as I usually work directly on the CSS file. I have the SCSS for other developers who might be interested in playing with it.

d13:
@alvarotrigo - Yes, the `sass` task compiles the SCSS to `.css`. If you're providing a SCSS file, you should provide the means to build it in the way you expect the final CSS to look. 

What I suggest as middle ground is to remove not run in the default task and just have the developer run the sass task manually before running the compression tasks.

alvarotrigo:
Yep, removing it from the global task should do it as I wouldn't personally use it.

Issue #1489, opened 2015-09-08
Auto-height not working with `scrollBar:true` and `fitToSection:true`


jojomcbobo:
Hello, thanks so much for this plugin it's amazing! I've been trying for a couple of days to get a footer section working but I'm getting some odd behavior I can't fix: http://www.jonasnaimark.com/temp/fullscreentest/Test.html#page12

The page will sometimes auto scroll up one section, other times it will stay but the nav will go back up after a delay. Often when hitting up from the footer it will jump up 2 sections instead of one. Any tips on how to fix this?

Thanks!

alvarotrigo:
It seems to be related with the use of `scrollBar:true`.
Thanks for reporting it. I'll investigate it.

alvarotrigo:
As a workaround, you can use `fitToSection:false` to prevent the problem.

alvarotrigo:
Fixed in fullpage.js 2.7.0

jojomcbobo:
Wow thanks for the update! It still seems to be happening on 2.7.0 though if scrollBar:true. Changing fitToSection doesn't fix it either. It works fine if scrollBar:false.

Updated to 2.7.0 with the same bug:
http://www.jonasnaimark.com/temp/fullscreentest/Test.html#page12

alvarotrigo:
Right. I'll take a look with more calm for the next release. 
However it doesn't seem to be a very simple problem to solve.

alvarotrigo:
> Changing fitToSection doesn't fix it either. 

Not sure to understand you there. 
Using `fitToSection:false` should fix the problem.

jojomcbobo:
Thanks again for looking at it!

rootkowski:
Not sure this will help in your case, but I needed to disable fitToSection on a project but it would still kick in on resize, rotate and sometimes on nav click (the section would be refitted over and over despite autoScrolling and fitToSection set to false). My workaround for that was to set fitToSectionDelay to 3600000 which I believe is one hour in milliseconds. That way the current section doesn't scroll to fill the screen, it stays where you want it, which is very good when you have a lot of content in sections and want to be able to view it all on a phone.

alvarotrigo:
@rootkowski update to the latest version 2.7.4.

rootkowski:
Thanks a lot for the super fast reply! :) 
This has definitely fixed it. Tested after removing the delay option.

jojomcbobo:
[...]

PR #1596, opened 2015-10-14
Fix scrolling back up for auto height element at bottom of page 

MariusRumpf:
As mentioned in #1489 
Bug shown here: https://jsfiddle.net/f4m48vbu/1/

Sources build with gulp as mentioned in README.
Tested in Chrome 45, Safari 9, Firefox 41, IE9.


alvarotrigo:
> // check if the end of the page is reached, needed for auto height element at bottom of the page

`fp-auto-height` can be applied in any section. It doesn't have to be the last one.

MariusRumpf:
This is a dirty fix then and shouln't be merged. My assumption was that this bug only occurs if the `fp-auto-height`element is at the bottom, since this is an exception in 
> // taking the section which is showing more content in the viewport

I will not find the time to rethink that logic then.

alvarotrigo:
No worries. The problem is not that urgent anymore. It can easily be fixed by using `fitToSection:false`.

PR #1610, opened 2015-10-23
Move fp-responsive class to the body tag

cedricblondeau:
Thus, we can easily apply CSS rules to elements outside the plugin container.

Can be particularly useful for hiding static positioned elements such as vertical navigation arrows (#814) which cannot be easily included in the plugin container because of the transform property.

Furthermore, `$(SECTION_NAV_SEL).hide();` and `$(SECTION_NAV_SEL).show();` in `setResponsive()` function could be replaced with a simple CSS rule:
`
.fp-responsive #fp-nav {
  display: none;
}
`


alvarotrigo:
Thanks for it!
Can you commit the change to the `dev` branch instead ?

Issue #728, opened 2014-09-22
Velocity.js integration


unaszplodrmann:
Hey :)

Thanks for the script. I've been having quite a lot of fun playing with it. Replacing the CSS3 transformations with the Velocity.js equivalent appears to work rather well. I considered using the transitionEnd event to release GPU resources after scrolling to a section/slide, but Velocity.js handles this automagically and is also backwards compatible with IE8/9. Given that it outperforms jQuery animations across the board and does not require CSS3 transform3d detection, it seems like a perfect fit for fullPage.js, apart from the fact that adds another dependency. I've offset this by using a non-standard jQuery build without the animation module etc.

Might it be worth looking at using Velocity.js instead of CSS3 or jQuery for animation?

elizabeth268:
Hi Alvaro,

Really interested in your plugin and very excited to integrate in my projects. Anyhow, I am more familiar with velocity and would like to use your plugin with it integrated in. In the post you've mentioned earlier:

> I believe I might integrate it as an option instead of using jQuery animations.

But I can't seem to find where the integration is in your code. Where might or how might I use the integration you've mentioned?

 Thank you. Amazed and truly appreciated by the work you've done for the community.

alvarotrigo:
@elizabeth268 if you read the whole topic you will notice that I said:
> I'm not sure to what extend does it make sense to integrate it in fullpage.js.

That's after I pointed out that I see not benefit at all in using velocity for fullpage.js or even for pagepiling.js. The performance is slower and therefore it won't be integrated.

If you prefer not to use jQuery, you can start to play with the [pure Javascript Alpha version of fullpage.js](https://github.com/alvarotrigo/fullPage.js/tree/master/pure%20javascript%20(Alpha)).

elizabeth268:
Thanks Alvaro. Looking at it right now and trying to play around with it. Anyhow, seeing that you are not to support it currently... makes me wonder if the pure Javascript Alpha version is slower in web performance? 

Also, as I am inserting your plugin into my project I noticed the css files that comes along. Are they required? But I thought this was a javascript plugin. How does the css come in?

alvarotrigo:
> makes me wonder if the pure Javascript Alpha version is slower in web performance?

The performance of the pure javascript version is even faster than the jQuery one when using the option `css:true`. Not sure about when using `css3:false`, but in any case, you should be using with `css3:true`.
[...]

PR #1622, opened 2015-10-29
add velocity.js support

lukasbesch:


alvarotrigo:
The topic [was treated time ago](https://github.com/alvarotrigo/fullPage.js/issues/728).

I didn't see any improvement using velocity comparing it to css3 animations. Therefore I don't think I'll consider this pull request.

PR #1628, opened 2015-11-01
added 'Sintelle' to 'Who is using fullPage.js' list

manioube:
we would really appreciate if you accepted our link


alvarotrigo:
Unfortunately I stopped adding new sites time ago and I only do it now with very original ones or the ones belonging to important companies.

manioube:
ok, thank you anyway

PR #1637, opened 2015-11-08
updated README.md

jarifibrahim:
The README.md file had 	

$('#fullpage').fullpage({
			menu: false
});

It should be
 menu: '#menu'


alvarotrigo:
Yeah, it could be that way as an example of usage. Thanks.

PR #1653, opened 2015-11-17
Added paddingTop and paddingBottom exceptions.

pedroadame:
Use 'fp-no-padding-top' and 'fp-no-padding-bottom' classes in a section to avoid paddingTop and paddingBottom initializer options respectively in that section.


pedroadame:
Why was this closed?

alvarotrigo:
Sorry for not givin an explanation. I though it was quite old for that :D

The reason is you can easily achieve this by yourself by just using CSS:

```css
.no-padding{
    padding-top: 0 !important;
}
```

pedroadame:
I already knew that, but somewhere I read using `!important` is a bad CSS practice.

alvarotrigo:
> I read using !important is a bad CSS practice.

Not in this case.

pedroadame:
Thank you for your advice then!

PR #1660, opened 2015-11-18
reading Window height based off of window.innerHeight vs jQuery's height method

rodriguez-alex:
updated to read window.innerHeight as it reliably delivers height of viewport even on devices where the chrome of the browser may hide after scrolling or on page rotation. fallback to jquery's height method if window.innerHeight is not supported

### DIFF
--- a/jquery.fullPage.js
+++ b/jquery.fullPage.js
@@ -381,7 +381,7 @@
             });
 
             var windowsWidth = $window.width();
-            windowsHeight = $window.height();  //updating global var
+            windowsHeight = window.innerHeight || $window.height();  //updating global var
 
             //text resizing
             if (options.resize) {
@@ -461,7 +461,7 @@
         var isTouchDevice = navigator.userAgent.match(/(iPhone|iPod|iPad|Android|playbook|silk|BlackBerry|BB10|Windows Phone|Tizen|Bada|webOS|IEMobile|Opera Mini)/);
         var isTouch = (('ontouchstart' in window) || (navigator.msMaxTouchPoints > 0) || (navigator.maxTouchPoints));
         var container = $(this);
-        var windowsHeight = $window.height();
+        var windowsHeight = window.innerHeight || $window.height();
         var isResizing = false;
         var isWindowFocused = true;
         var lastScrolledDestiny;
@@ -504,7 +504,7 @@
             FP.setAllowScrolling(true);
 
             //due to https://github.com/alvarotrigo/fullPage.js/issues/1502
-            windowsHeight = $window.height();
+            windowsHeight = window.innerHeight || $window.height();
 
             FP.setAutoScrolling(options.autoScrolling, 'internal');
 
@@ -978,7 +978,7 @@
                     else if(options.autoScrolling){
 
                         //is the movement greater than the minimum resistance to scroll?
-                        if (Math.abs(touchStartY - touchEndY) > ($window.height() / 100 * options.touchSensitivity)) {
+                        if (Math.abs(touchStartY - touchEndY) > ((window.innerHeight || $window.height()) / 100 * options.touchSensitivity)) {
                             if (touchStartY > touchEndY) {
                                 scrolling('down', scrollable);
                             } else if (touchEndY > touchStartY) {
@@ -1180,7 +1180,7 @@
 
         //IE < 10 pollify for requestAnimationFrame
         window.requestAnimFrame = function(){
-            return window.requestAnimationFrame || 
+            return window.requestAnimationFrame ||
                 window.webkitRequestAnimationFrame ||
                 window.mozRequestAnimationFrame ||
                 window.oRequestAnimationFrame ||
@@ -1763,7 +1763,7 @@
 
                 //if the keyboard is NOT visible
                 if (!activeElement.is('textarea') && !activeElement.is('input') && !activeElement.is('select')) {
-                    var currentHeight = $window.height();
+                    var currentHeight = window.innerHeight || $window.height();
 
                     //making sure the change in the viewport size is enough to force a rebuild. (20 % of the window to avoid problems when hidding scroll bars)
                     if( Math.abs(currentHeight - previousHeight) > (20 * Math.max(previousHeight, currentHeight) / 100) ){
@@ -1792,8 +1792,8 @@
 
             //only calculating what we need. Remember its called on the resize event.
             var isBreakingPointWidth = widthLimit && $window.width() < widthLimit;
-            var isBreakingPointHeight = heightLimit && $window.height() < heightLimit;
-            
+            var isBreakingPointHeight = heightLimit && (window.innerHeight || $window.height()) < heightLimit;
+
             if(widthLimit && heightLimit){
                 FP.setResponsive(isBreakingPointWidth || isBreakingPointHeight);
             }


alvarotrigo:
Can you name what's the advantage of using innerHeight and elaborate the explanation ?

alvarotrigo:
In any case, I would use a function to get the value:

```javascript
getWindowHeight(){
    return window.innerHeight || $window.height();
}
```

Can you commit these changes to the DEV branch instead?

rodriguez-alex:
The advantage we found was that the window.innerHeight gets the true viewport size even after the chrome disappears in some browsers (such as iOS) compared to other options of grabbing the height. We ran a test on iOS and window.innerHeight gave us the extended height (after the chrome was gone) vs the other options that gave us the height as if the chrome was still there. The other ways of retrieving height we attempted were:

$(window).height();
$('body').height());
$('html').height());
document.documentElement.clientHeight;
document.body.clientHeight;
$(document).height();

I have committed the changed into the dev branch now.

alvarotrigo:
> even after the chrome disappears in some browsers (such as iOS)

What do you mean with the "chrome"? The address bar? Why would the address bar disappear ? 
What fullpage.js options are you using? scrollBar:true? autoScrolling:false?

rodriguez-alex:
Chrome, meaning the browser sections (bookmark bar, address bar). We ran into an edge case where if we turn on autoScrolling true (allowing the user to scroll) then turn it back off, the chrome loses the bookmark bar on iOS if they scroll. We tried rebuilding when a user proceeded to the next slide, however all height options gave us the same height as if the bookmark bar was still there so the height was still messed up. The only one that gave us a true height was window.innerHeight. Videos below.

Before using $(window).height:
http://screencast.com/t/ZC01Y50i3

After using window.height: 
http://screencast.com/t/K1V4Z7ivotu

rodriguez-alex:
Woops, didn't know deleting it would close this request.

alvarotrigo:
@rodriguez-alex there's something I don't understand about your videos. On iPhone, whenever you scroll up, the address bar (at the top) and the actions bar (at the bottom) will become visible again.
(Tested on iOS 8 and the latest 9)

I was not able to see that in your videos. Maybe you are using the mouse wheel rather than the finger (pressing with the mouse) to scroll?

The innerHeight solution that you propose would hide part of the site behind the bottom action bar.

rodriguez-alex:
Hey @alvarotrigo, the reason you see the scrollbar come up is because you are swiping up rather quickly (we've done this test multiple times). If you scroll up slowly, then you will see that it stays disappeared. Us developers have gotten used to just rolling through things, ha!

If you scroll down a little, then scroll up slowly, say if a user is reading or is trying to see details of something, OR if they scroll down a bit and notice they don't like the content and scroll back up then the bar will stay hidden.

I recommend we also implement a check when the user switches between slides to verify the fp-section is the same as the innerHeight. As of now, we've manually added this into the afterLoad callback. This should be baked into the plugin. I'll add it into the develop pull request. Apologies on adding it to master here, first time doing a pull request :) Will update in an hour or two.

alvarotrigo:
I'm not able to reproduce your problems [in the example page](http://alvarotrigo.com/fullPage/examples/normalScroll.html) using `autoScrolling:false`.

> If you scroll up slowly, then you will see that it stays disappeared.

Not in my iPhone 5 iOS 8 (and neither in the testings I did in browserstack.com) No matter how slowly you scroll up, the bars will appear, but slower. 


> If you scroll down a little, then scroll up slowly,

I wasn't able to reproduce it either.

alvarotrigo:
> I recommend we also implement a check when the user switches between slides to verify the fp-section is the same as the innerHeight.

That would only execute once per section, but in the case of going to the 1st section, as I said, the top and bottom navigation bars appear bit by bit at the same time of the scroll and at the same speed. 
So this wouldn't work for it.

rodriguez-alex:
I took video of your site on the simulator and actual device. If you scroll away from the section off for a bit it disappears. You can see there is no re-rendering of the section when the chrome disappears. 

Simulator:
http://screencast.com/t/tUoUxNJg 

Actual device
https://drive.google.com/file/d/0B_Qvluhviq3HR2RaRURMRS1zSHM/view?usp=sharing

We've implemented the solution in "afterLoad", since it runs after the scrolling has ended. At this point, we update all 'fp-sections' the match the window.innerHeight using jquery.

alvarotrigo:
I'm talking about returning to the 1st section.

rodriguez-alex:
First section:
http://screencast.com/t/EiNmKWdWfSpf

alvarotrigo:
It seems that's being fired by the `fitToSection` functionality. Not by your finger swiping the screen. 
Right?

rodriguez-alex:
It seems it is. I haven't completely looked at every functionality of your code, but are you snapping to each section by looking at the anchors on the page (we have `lockAnchors` set to false)?

Nevertheless, either swipe or auto-scroll to section, we would want the height to be the same correct? :)

Here's our settings:

```
css3: false,
      lockAnchors: true,
      scrollingSpeed: 1000,
      scrollBar: true,
      fitToSectionDelay: 100,
      normalScrollElements: '.menu_mainNav',
      keyboardScrolling: false,
```

Glad you're being thorough w/ your code and on top of asking questions, you must be a good developer :)

alvarotrigo:
> we would want the height to be the same correct?

Ideally, but it seems it not easy to please both cases:
- First section full screen without hiding anything when swiping to it
- Rest of sections full screen without showing other section's content

I use to recommend using fullpage.js without anchors, which will keep both the top and bottom bar visible on iPhone devices when using `autoScrolling:true`, but I know that's not your case.

rodriguez-alex:
- The only time the first section would hide when swiping into it, would be if scrolling was enabled (as in our case), but it would then be corrected by the fp-section height vs window.height check. 

- The rest of the sections would also only once, show part of the next section's content, but automatically corrected once fp-section vs window.height check is done as well.

What do you think? :) I really want to get the window.innerHeight in lol

alvarotrigo:
> it would then be corrected by the fp-section height vs window.height check.

I believe the callback `afterLoad` will get fired before the top and bottom bar show themselves completely in some cases (slow swipe up for example). Feel free to test it by yourself. 

In any case, I would really like to see how this behaves in other devices with other OS (Android, Windows Phone..). What might work for iOS might not be ideal for others and dealing with address bar and sizes is always a problem.

If you modify fullpage.js according to your indications, please upload it somewhere so we can test it in different devices before taking any decision.

rodriguez-alex:
Yeah, I'm looking at it now in your demo, in your `afterSectionLoads` it acts a lil wonky, will take some time especially w/ all other scenarios. How about we test window.innerHeight vs $(window).height in the meantime? :)

alvarotrigo:
> How about we test window.innerHeight vs $(window).height in the meantime? :)

Yeah sure, feel free to test it in your own fullpage.js version.

rodriguez-alex:
What devices/browsers/scenarios did you see issues in initially that I should check that they still work?

alvarotrigo:
I haven't tested it at all, as I still concerned by the problem of the 1st section and the top and bottom bars.

But it would be nice if you could test it in other devices as well before going forward with this idea.

rodriguez-alex:
I've made dev request here: https://github.com/alvarotrigo/fullPage.js/pull/1690

PR #1677, opened 2015-11-25
Fix window width bug

artifex404:
When fullpage.js has a window width trigger, the responsiveWidth setting it does not perform as it should.

$(window).width() does not take window scrollbar into consideration, while CSS media queries do. So if the page is responsive, CSS media queries and fullpage.js does not perform in sync. As different browsers have a different pixel amount for the scrollbar width, the only option is to use the window.outerWidth to perform calculations.


alvarotrigo:
Can you commit the changes to the `dev` branch instead?

artifex404:
Sure, I'll change the branch.

artifex404:
Moved to dev branch. Closing this pull request.

PR #1678, opened 2015-11-25
Window width bug

artifex404:
When fullpage.js has a window width trigger, the responsiveWidth setting it does not perform as it should.

$(window).width() does not take window scrollbar into consideration, while CSS media queries do. So if the page is responsive, CSS media queries and fullpage.js does not perform in sync. As different browsers have a different pixel amount for the scrollbar width, the only option is to use the window.outerWidth to perform calculations.


alvarotrigo:
I probably agree with the outerWith. But not quite sure about the outerHeight. It might create problems with mobile browsers and their top and bottom bars.

alvarotrigo:
Take a look at [this is a similar topic](https://github.com/alvarotrigo/fullPage.js/pull/1660) in an attemp to change the `$(window).height();` for `window.innerHeight`.

artifex404:
Agree. I'll revert the window.outerHeight to the jquery function $window.height() for a new pull request.

alvarotrigo:
Thanks for it! I'll test it before merging it with the next release.

alvarotrigo:
Merged in 2.7.6

alvarotrigo:
This leads to problems when using fullpage.js inside an iframe, a clear example of it can be found in [jsfiddle](http://jsfiddle.net/97tbk/1007/) or codepen. 
The responsive options won't fire on them because it is calculating the width of the main page instead of the iframe.

I suggest some check like:

```javascript
var windowsWidth = container.closest('iframe') ? $window.width() : window.outerWidth;
```

alvarotrigo:
Seems to be giving problems with phones... I'll probably have to roll it back.

PR #1679, opened 2015-11-26
Master

oimken:
Fixed: Scroll failing when first section be set to auto height（after
scroll down to other section,  can not scrolling up to first section,
first section will not be display again . ）


alvarotrigo:
Thanks for it! 

Two things:

- Convert it to this function:
```javascript
        function getDestinationTop(dest, element){
            //auto height? Scrolling only a bit, the next element's height. 
            if(element.hasClass(AUTO_HEIGHT) && dest.top){
                return dest.top - windowsHeight + element.height();
            }

            //Otherwise the whole viewport.
            return dest.top;
        }
```

And get it like this:
```javascript
var dtop = getDestinationTop(dest, element);
```

- And commit it to the `dev` branch instead.

alvarotrigo:
Or you can even it it in a single line:

```javascript
var dtop = element.hasClass(AUTO_HEIGHT) && dest.top ? (dest.top - windowsHeight + element.height()) : dest.top;
```

alvarotrigo:
Why did you close it?

oimken:
sorry for closed it. I am not a professional coder.  and not  familiar with github ... really sorry for that. I will do better next time :)

PR #1680, opened 2015-11-26
Fixed: Scroll failing when first section be set to auto height

oimken:
Fixed: Scroll failing when first section be set to auto height（after
scroll down to other section,  can not scrolling up to first section,
first section will not be display again . ）


alvarotrigo:
Why did you close it?

oimken:
I edited package.json also, I found it after sent pull request. sorry . 
this fix is tooooo simple and I feel it is not elegant enoght to be pull...
I want to post this bug on issues instead pull request, but forgot.

btw: thanks for your sharing  this great work.  I love it.

PR #1682, opened 2015-11-28
Fixed: Scroll failing when first section be set to auto height

oimken:
Fixed: Scroll failing when first section be set to auto height
After scroll down to other section, can not scrolling up to first section,
first section will not be display again .

https://github.com/alvarotrigo/fullPage.js/pull/1679


alvarotrigo:
Thanks for it!

PR #1690, opened 2015-12-01
Dev

rodriguez-alex:
I've updated the fullpage.js to use window.innerheight, and have created a check to see when the page has stopped moving to ensure it's only called once. I have tested this out in Iphone 6+, Samsung S6, Nexus 6. Here's the URL:

http://test2.forgedpixels.net/examples/normalScroll.html


alvarotrigo:
Thanks for it @rodriguez-alex, although I'm not yet totally convince about it.
If you try to scroll up or down while the sections are about to resize, then the scrolling won't take place  and some glitches will appear.

rodriguez-alex:
What do you mean?

My end goal is to get window.innerHeight in there, lol. I do like the extra built in functionality to make sure it fits the screen though :)  Had to do some thinking for sure.

alvarotrigo:
> What do you mean?

Try to scroll up or down just when the resize is about to take place. You'll see the scroll won't take place and the user will be forced to stay in the same section until the resize has finished.

rodriguez-alex:
We can clear the timeout when they're scrolling if the rebuild is the issue. Is this what you mean?

alvarotrigo:
It shouldn't interrupt the user input.

rodriguez-alex:
I've updated the demo with the new changes, looks good now :dancers:

DomenicRoti:
Hey, I just tried this out and didn't see any problems, any chance we roll this in?  Thanks

alvarotrigo:
@DomenicRoti chances are, but I still need to take a look at it with more calm. It won't be merged before January.

alvarotrigo:
I'll close the PR temporally.

rodriguez-alex:
Hey @alvarotrigo any updates on this? Do we see a merge anytime soon? Thanks in advance.

alvarotrigo:
Probably for the next release. 
Could you please update it to avoid conflicts with the current version 2.7.7 ?

alvarotrigo:
And get rid of the `normalScroll.html` file, it doesn't need to be modified to test it as far as I understood.

rodriguez-alex:
I will update it today.

PR #1704, opened 2015-12-07
fix potential problem

t-ashula:
Fix potential problem with old jQuery and newer jQuery with Migrate plugin.


alvarotrigo:
Can be you be more specific with the problem it solves?
Also, commit it to the dev branch instead.

t-ashula:
reopened as #1705

PR #1705, opened 2015-12-08
fix potential problem

t-ashula:
Fix the potential problem with old jQuery and newer jQuery with Migrate plugin.

i don't want to write specific about this problem because of it's a sort of security issue.
see https://github.com/jquery/jquery-migrate/issues/36 and https://bugs.jquery.com/ticket/9521


alvarotrigo:
Got it :) Thanks a lot for it!

alvarotrigo:
Merged in 2.7.6

PR #1752, opened 2016-01-06
Fix the wrong active section flag under normal scroll

hang-qi:
Under normal scroll (`autoScrolling: false`), when a section is longer than the window height, the active flag in navigation can be wrong. This PR fixes this issue.

The current logic (in the base fork) calculates the difference between the `section.offsetTop` and `$window.scrollTop()` for each section and pick the one with the smallest diff as active. However, this can be wrong. For example, suppose there are two sections A and B, where section A is much longer than the window height. Consider the case when you have *just* scrolled B entering the viewport. At this point, section B occupies very little area of the window, whereas section A still dominates. However, `diff(A.offsetTop, window.ScrollTop) < diff(B.offsetTop , window.ScrollTop)` as A is very long and the top of B is much closer to the window scrollTop. This will select section B as the current active section *wrongly*.

To fix this issue, I changed the logic to explicitly find _which section passes the middle line of the screen_. This will work correctly for the scenario above, where the top of section B obviously does not pass the middle line. 

Corner cases:
* If section B is short and there is a section C coming, selecting section B as active when it passes the middle line is still acceptable because it is the section at the screen center (which is the center of attention).
* In the case of `autoScrolling: true`, my modification works as well. So there is no negative side-effect.

### DIFF
--- a/jquery.fullPage.js
+++ b/jquery.fullPage.js
@@ -790,18 +790,17 @@
             if(!options.autoScrolling || options.scrollBar){
                 var currentScroll = $window.scrollTop();
                 var visibleSectionIndex = 0;
-                var initial = Math.abs(currentScroll - document.querySelectorAll(SECTION_SEL)[0].offsetTop);
+                var screen_mid = currentScroll + ($window.height() / 2.0);
 
                 //taking the section which is showing more content in the viewport
                 var sections =  document.querySelectorAll(SECTION_SEL);
                 for (var i = 0; i < sections.length; ++i) {
                     var section = sections[i];
 
-                    var current = Math.abs(currentScroll - section.offsetTop);
-
-                    if(current < initial){
+                    // Pick the the last section which passes the middle line of the screen.
+                    if (section.offsetTop <= screen_mid)
+                    {
                         visibleSectionIndex = i;
-                        initial = current;
                     }
                 }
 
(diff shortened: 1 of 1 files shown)


alvarotrigo:
Thanks for it @hang-qi! 
I'll take a look at it. 

fullpage.js didn't permit the use of bigger sections than the viewport in the past, so that's why this problem was there.

alvarotrigo:
> If section B is short and there is a section C coming, selecting section B as active when it passes the middle line is still acceptable because it is the section at the screen center (which is the center of attention).

To do so you would have to change in line 813 this:

```javascript
if(!currentSection.hasClass(ACTIVE) && !currentSection.hasClass(AUTO_HEIGHT)){
```

For this
```javascript
if(!currentSection.hasClass(ACTIVE)){
```

Otherwise sections with `fp-auto-height` wont get active (and the menu). You can test it by using the `normalScroll.html` example and adding the `fp-auto-height` to the 2nd section.

We still have troubles when using `scrollBar:true` in this cases, as you can scroll to the 2nd section but this won't pass the middle of the screen. Although I believe that's for another PR.

alvarotrigo:
If you want to add the change I'll merge the changes after it.
Otherwise I'll merge it anyway and apply the change later on.

hang-qi:
Thanks for the comment! I've included the change you mentioned.

alvarotrigo:
Thanks for it!

Issue #1727, opened 2015-12-22
IE11 does not scroll


Riggi:
Hi,

Thank you for your great plugin - it helped me a lot!! I really appreciate your work! It's made in a smart way so i can really do what I wanted to with it.

I'm using it on this site http://www.ipeak.ch/tourismus-communicator/ and it's working well on all browsers but the IE11. I really tried googling the error but I can't find anything helpful.

The onLeave and afterLoad events do fire, but the 'big frame' does not slide.

Is there any way for a workaround?

Thanks for your time

Rick

alvarotrigo:
The [demo page](http://alvarotrigo.com/fullPage/) works on IE 11. You will have to provide an isolated reproduction in jsfiddle or codepen if you believe its a bug in fullpage.js. 

I'll close the issue until then.

panther7:
I have same problem, 2.7.5 works, 2.7.6 does not :-/

alvarotrigo:
@panther7 provide an isolated reproduction of the issue and specify which device and browser you are using.

panther7:
In 2.7.5 does not works too, i think that problem with MouseWheelHandler(), i will try reproduction tomorrow.

panther7:
#1778

panther7:
Bug reproduced... try add CSS overflow(-x): hidden to "fullpage-wrapper".

alvarotrigo:
> Bug reproduced... try add CSS overflow(-x): hidden to "fullpage-wrapper".

And why would you add `overflow-x: hidden`... It seems to me that's not a problem of the plugin but yours with the plugin and your setup.

panther7:
Yes, problem is in finally implementation (css), but i did reproduced the problem, which is on the http://www.ipeak.ch/tourismus-communicator/ too. ;-)

panther7:
But, stupid is, that works in FF and Chrome, not in IE11 (maybe lower)

alvarotrigo:
I don't see why would you  need `overflow:hidden`

panther7:
Easy, development, in past was useful, but after time, it was forgotten.

PR #1778, opened 2016-01-20
fix for IE11

panther7:


alvarotrigo:
You will have to be more specific with the problem you are fixing. 

[The demo page](http://alvarotrigo.com/fullPage/) works as expected in IE 11.

alvarotrigo:
That's not enough. 
You didn't provide any way of reproducing it and it is not clear what are you trying to solve. 

As I said previously:

> @panther7 provide an isolated reproduction of the issue and specify which device and browser you are using.

panther7:
#1727

alvarotrigo:
I'll close it as I see no need for it.

PR #1843, opened 2016-02-22
add offset for top

matgargano:
I ran into an issue where scrolling to sections wasn't working as I had wanted to because of other elements on the page... (see screenshot) I added a functional callback (also accepts an integer) for the value offset... functional parameters allow for on the fly calculations (e.g. different offsets at different breakpoints)...

screenshot 2016-02-21 21 49 36

the way it can be used is:

```
  $('#fullpage').fullpage({
        menu: '#menu',
        autoScrolling: false,
        fitToSection: false,
        offset: function () {
            var heightOffset = -60;
            if ($(document).width() < 1000) {
                heightOffset = -10;
            }
            return heightOffset;
        }
    });
```

or

```
  $('#fullpage').fullpage({
        menu: '#menu',
        autoScrolling: false,
        fitToSection: false,
        offset: 20
    });
```


alvarotrigo:
Not quite sure what's the purpose of it. Can you explain your situation in another way? (and ideally with an example in jsfiddle or codepen)

The plugin already provides `paddingTop` and `paddinBottom` options for cases in which you want to use a header or a footer. Any element outside the plugin's wrapper has to be fixed positioned. 

I'll close the pull request anyway  because as detailed [in the docs](https://github.com/alvarotrigo/fullPage.js/wiki/Contributing-to-fullpage.js), pull requests should be made to the `dev` branch.

matgargano:
Your padding top and bottom didn't work. This was the only way. I'll recreate the PR on Dev.

alvarotrigo:
> Your padding top and bottom didn't work.

You would have to provide a reproduction proving that. [It seems to work as expected](http://jsfiddle.net/97tbk/1074/).

Issue #1854, opened 2016-02-25
Don't want to scroll to the bottom of a section


NiklasLazinbee:
Hallo,
for mobile divices with small screens I set `$.fn.fullpage.setAutoScrolling(false)`.
I have sections in which the height is more than screenheight, when I use the menu to skip to a section, fullPage.js animates to the bottom of these section. But I want to go to the top of  the section. I don't want to use `$.fn.fullpage.setFitToSection(true)` because than I at first go to the bottom and than the script jumps to the top of the section.

Is there a way that fullPage.js animates to the top of a section and not to the bottom?

Best regards
Niklas

alvarotrigo:
> Is there a way that fullPage.js animates to the top of a section and not to the bottom?

fullPage.js will animate to the top if you come from a section above the destination one.
It will animate to the bottom if you come from a section below the destination one.

Make sure you use the latest fullpage.js (2.7.8) to get this result.

NiklasLazinbee:
I find a solution which works for me, but I don't fi it works for everyone because I don't know all dependencies:
I change line 1294 from
`if(element.outerHeight() > windowsHeight){`
to 
`if(element.outerHeight() > windowsHeight && options.autoScrolling){`
and line 1302 from
`else if(isScrollingDown){`
to 
`else if(isScrollingDown && options.autoScrolling){`

Beste regards 
Nikls

pebarrault:
Hi,

I'm in the same case despite using the 2.7.8 version. 

This fix works for me, thanks !

dceonline:
Hi,

that works for me too. Maybe you can add an option like 'scrollToSectionTop' true|false, to always scroll to the top off a section. That would be a nice feature if autoscrolling is disabled and someone uses the navbar.

Regards
Ortwin

codercooooder:
I'm not sure if I implemented this correctly, but as I see it this fix requires you turn off autoScrolling?  It is an alright fix because it preserves the menu functionality, but for me I lose all scrolling animation (which I'd like to retain).  I woudl love it if someone could come up with a fix that would allow me to keep autoscrolling on and instead scroll to the top from section when scrolling up?

Klasyczny:
Hello, 
while solution provided in 2.7.8 and proposed here by Niklas works fine on a computer or google emulated devices, it doesnt work on mobile devices. There, skipping to section still jumps to bottom instead of top of it. I hope this isn't the impassable problem and can be handled.

Regards,
Klasyczny

[...]

PR #1867, opened 2016-03-01
Add support for custom getDestinationPosition

nadavspi:
Allow the user to optionally pass a custom `getDestinationPosition` handler. 
My use case is something like this:

```javascript
    $('#fullpage').fullpage({
        getDestinationPosition: function(dest, element) {
            // Position the element in the middle of the viewport if there's room,
            // or at the top if there isn't.
            var elementOffset = element.offset().top;
            var elementHeight = element.height();
            var windowHeight = $(window).height();

            var position = elementOffset;
            if (elementHeight < windowHeight) {
              position = elementOffset - (windowHeight / 2 - elementHeight / 2);
            }

            return position;
        }
    });
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,7 +198,8 @@ $(document).ready(function() {
 		afterRender: function(){},
 		afterResize: function(){},
 		afterSlideLoad: function(anchorLink, index, slideAnchor, slideIndex){},
-		onSlideLeave: function(anchorLink, index, slideIndex, direction, nextSlideIndex){}
+		onSlideLeave: function(anchorLink, index, slideIndex, direction, nextSlideIndex){},
+ 		getDestinationPosition: function(dest, element){}
 	});
 });
 ```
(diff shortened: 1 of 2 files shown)


alvarotrigo:
Thanks for it, but I believe I'll opt for a much simpler solution, something like what @dceonline proposed in https://github.com/alvarotrigo/fullPage.js/issues/1854, an option like `scrollToSectionTop` or similar.

nadavspi:
OK. Your call, but note that `scrollToSectionTop` doesn't cover my use case (and potential future use cases that aren't just top or bottom.)

toddbc:
That doesn't allow handling the use case of scrolling so that the content is centered.

I feel like this situation is much like animation easing functions: it's good and convenient to have nicknames for very common ones, but it's also nice sometimes to be able to define a custom curve.  If the presets are just alternate functions, it remains fairly simple, no?

alvarotrigo:
Are you guys talking about smaller sections than the viewport? Or larger sections?

nadavspi:
Smaller than the viewport. Here's an example: 
screenshot 2016-03-07 11 00 53

alvarotrigo:
Could you provide a jsfiddle or codepen so I can get a better idea of what your purpose is?
If not `top` or `bottom`, maybe a `middle` option could also be provided to solve that problem with smaller sections?

MaZderMind:
I'd have needed something similar.
My Customer wants Sections, each 50% of the Screen height.
The active Section should be centered vertically, while the prev. and next section should be visible 1/2.

I needed to patch getDestinationPosition to accomplish that.
It's always nice to be able to override such important functions without patching the original file or even waiting for a PR to come through.

Issue #1835, opened 2016-02-17
silentMoveTo still shows animation when scrolling to a slide


adriennetacke:
Hello! 

Please check out the quick fiddle: 
http://jsfiddle.net/zegm4xag/1/

Here, it still shows the animation instead of cutting straight to the requested slide. 

Is there something I am missing?

alvarotrigo:
Not missing anything. That's a bug.

adriennetacke:
Ah, ok. Thanks for validating!

alvarotrigo:
Do this meanwhile:

```javascript
$('.move').click(function() {
	$.fn.fullpage.setScrollingSpeed(0);
    $.fn.fullpage.moveTo(0, 3);
    $.fn.fullpage.setScrollingSpeed(700);
});
```

adriennetacke:
Perfect. Thank you.

nhylated:
Hi @alvarotrigo. I've fixed this in my dev branch [here](https://github.com/nhylated/fullPage.js/commit/f715d9c49c433f12199f82e056276fff95765b99). Can I create a pull request?

alvarotrigo:
@nhylated yep, you may.

nhylated:
@alvarotrigo Done - #1885 

Thanks!

alvarotrigo:
Fixed on the `dev` branch 2.7.9

alvarotrigo:
Fixed in 2.7.9

PR #1885, opened 2016-03-08
Fix for suppressing animation during silentMoveTo (#1835).

nhylated:
Fix for suppressing animation during silentMoveTo (#1835).

### DIFF
--- a/jquery.fullPage.js
+++ b/jquery.fullPage.js
@@ -345,12 +345,14 @@
         FP.silentMoveTo = function(sectionAnchor, slideAnchor){
             requestAnimFrame(function(){
                 FP.setScrollingSpeed (0, 'internal');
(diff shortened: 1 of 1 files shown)


adriennetacke:
Just wondering when I can expect this pull request to be merged. The sooner, the more appreciative I would be. :)

Thank you! Especially you, @nhylated for the fix!

alvarotrigo:
I didn't merge it yet as I'm thinking of removing the `requestAnimatinoFrame` calls everywhere. 
They were introduced to deal with [a Chrome bug](https://github.com/alvarotrigo/fullPage.js/issues/1389) time ago, but they may no longer be necessary.

alvarotrigo:
Fixed in 2.7.9.

PR #1896, opened 2016-03-10
fix the jquery issue, when we use the new version jquery

bangwu:


alvarotrigo:
Thanks for it!
But, as detailed in the [contributing to fullpage.js indications](https://github.com/alvarotrigo/fullPage.js/wiki/Contributing-to-fullpage.js), can you please **Describe the problem in the Pull Request description**

What is *the jquery issue**? How to reproduce it? Under which environment? Under which initialization?
Providing a jsfiddle or codepen with the reproduction would be ideal.

bangwu:
OK, Thanks, I got it, I will follow the step.

alvarotrigo:
Also, in case of adding the pull request, it will have to be in the `dev` branch as stated in the contribution section too.

PR #1898, opened 2016-03-10
fix the old version jquery

bangwu:
When we use the newest jquery version, it will occur this issue"Uncaught referenceError: JQuery is not defined"


alvarotrigo:
Can you please provide a reproduction of the issue? I can not reproduce it here in [this fiddle](http://jsfiddle.net/97tbk/1109/).

alvarotrigo:
Closed as no reproduction was provided.

PR #1927, opened 2016-03-26
Add staticUrl option

jraisanen:
Added a static url option, to give the possibility to leave the url unchanged when switching between sections and slides.


alvarotrigo:
There's an option for that already, it is called `lockAnchors`.

PR #1961, opened 2016-04-11
Change default 'active' class to 'fp-active'

netes:
Hey,
I'm wondering... wouldn't be better to change default `active` class to `fp-active` to prevent namespace conflicts? For example with Bootstrap `active` class. I ran into this issue today and changed it.

P.S. Thanks for the great addon!

### DIFF
--- a/jquery.fullPage.js
+++ b/jquery.fullPage.js
@@ -35,7 +35,7 @@
     var DESTROYED =             'fp-destroyed';
     var ENABLED =               'fp-enabled';
     var VIEWING_PREFIX =        'fp-viewing';
-    var ACTIVE =                'active';
+    var ACTIVE =                'fp-active';
     var ACTIVE_SEL =            '.' + ACTIVE;
     var COMPLETELY =            'fp-completely';
     var COMPLETELY_SEL =        '.' + COMPLETELY;


alvarotrigo:
Can you specify where exactly is the conflict with bootstrap? Which element in bootstrap uses the active class that can be used at the same time for sections or slides?

Or is it only the active class in the navigation list element what conflicts with it?

If we are talking about the 2nd case, then we can create a new class for the navigation elements. Otherwise then you will have to wait for fullpage.js v3 for that. Changing the "active" class applied to sections and slides is a major change that can break many websites using fullpage.js. 
Those kind of changes are only applied on new versions.

netes:
It seems that this PR was a little premature, sorry for that. I made some investigation and it seems that the problem cause was custom `.active` class without any specificity. I changed that class and everything is fine now. But all in all - I think it would be nice to add option to customize `active` class name :)

alvarotrigo:
> it seems that the problem cause was custom .active class without any specificity.

I'm not able to see any CSS style applied by Bootstrap 3.3.6 in the active class alone.
You can verify it [in this jsfiffle](https://jsfiddle.net/fqg6zah8/) by using the dev tools.

> I think it would be nice to add option to customize active class name

I agree, for version 3 probably :)

netes:
> I'm not able to see any CSS style applied by Bootstrap 3.3.6 in the active class alone.
Yes, exactly. Someone added custom class directly to Bootstrap CSS files 😕

alvarotrigo:
Right I see.
I'll leave the PR open in case I merge it for v3.

netes:
Great, thanks!

BossElijah:
You could also make a prop for it where users can define their own custom active-class. That way, you can just keep the current class, and make people customize if they want, and it won't break any existing sites.

alvarotrigo:
> You could also make a prop for it where users can define their own custom active-class. That way, you can just keep the current class, and make people customize if they want, and it won't break any existing sites.

Yeah, that could be a good approach.

Issue #1852, opened 2016-02-24
Responsive auto-height


panczarny:
Hi,
how can I enable responsive auto-height (add class 'fp-auto-height')?
E.g. under 900px make each ".section" (or selected ones) auto-height.

Thanks!

alvarotrigo:
I would do some conditional CSS like this:

```css
.fp-responsive #contact-section,
.fp-responsive #contact-section .fp-slide,
.fp-responsive #contact-section .fp-tableCell{
    height: auto !important;
}
```

The `fp-responsive` class is one of [the state classes](https://github.com/alvarotrigo/fullPage.js#state-classes-added-by-fullpagejs) added by fullpage.js.

I'll mark it as an enhancement in any case. It is a nice feature to have.

alvarotrigo:
Merged on dev. Ready for the next release.

alvarotrigo:
Available now in 2.8.0.

codeevashum:
i have a page with two divs. div1 one with 10% of the height  and fixed, div2 with 90% height where i want the fullpage.js to run and fit the div without taking up the whole height of the window...  is there a way to fit it to a divs height ?

alvarotrigo:
Read [the docs](https://github.com/alvarotrigo/fullPage.js) and look for `fp-auto-height`.
You might want to look into the [Offset Sections](http://alvarotrigo.com/fullPage/extensions/offsetSections.html) extension.

trainoasis:
For those who used NPM to add fullpage.js: you have to manually import CSS in your sass if you use it, this works for me: 
```
/**
 * Import npm dependencies
 *
 * Prefix your imports with `~` to grab from node_modules/
 * @see https://github.com/webpack-contrib/sass-loader#imports
 */
@import "~fullpage.js/dist/jquery.fullpage.css";
```

kirkbross:
With jQuery:

`$(window).resize(function () {
    autoHeightMobile();
  });
  autoHeightMobile();

function autoHeightMobile() {
  var windowWidth = $(window).width();
  if (windowWidth < 768) {
    $('section.section').addClass('fp-auto-height');
  } else {
    $('section.section').removeClass('fp-auto-height');
  }
}`

alvarotrigo:
@kirkbross that's not necessary anymore as fullPage.js now provides the option `fp-auto-height-responsive`. 

Just use the `responsive` options and the class `fp-auto-height-responsive` in the sections that you need.

PR #1968, opened 2016-04-15
Added .fp-responsive-auto-height, fix for issue #1852

panczarny:
Fixing issue #1852

This is how it works with my changes:
http://codepen.io/panczarny/pen/mPxNyr

Adde things in .js, not in .css file.
Now just add ```.fp-responsive-auto-height ``` class to the ```.section ``` to enable its auto-height when in responsive.


alvarotrigo:
Thanks for it! Good idea.

Although it might be easier just to add the following in the CSS stylesheet:

```css
.fp-responsive .fp-responsive-auto-height.fp-section,
.fp-responsive .fp-responsive-auto-height .fp-slide,
.fp-responsive .fp-responsive-auto-height .fp-tableCell{
    height: auto !important;
}
```

panczarny:
I will also make adding ```.fp-responsive``` and ```.fp-responsive-auto-height``` independet, so some ```.section```s would become auto-height when fullpage is still working, e.g.
```
	$("#fullpage").fullpage({
		responsiveWidth: 750,
		responsiveAutoHeight: 1000
	});
```
But you're right, I'll take it into css.

panczarny:
As you said I moved it to css.
Added new option called ```responsiveAutoHeightWidth``` which makes my improvement independent.
Check this out at http://codepen.io/panczarny/pen/mPxNyr.

alvarotrigo:
Too many changes in the CSS file. The whole file changed. Make sure you use the same indentation as in the original file so I can clearly see the differences.

Regarding making responsive or responsive auto height independent doesn't make much sense to me, as usually when using `fp-auto-height` in this cases users want to create bigger sections, not smaller ones and fullpage.js won't deal properly with bigger sections than the viewport when using `autoScrolling:true`.

alvarotrigo:
If you still planning to commit it again with the changes I mentioned let me know. 
Otherwise I'll add them by myself.

panczarny:
Well, I didn't checked what happens if section is oversized with these functions independent. It's just like you said.
I will commit it with proper changes.

panczarny:
Commited.
Now added is only auto-height reponsive in css and scss files. Removed my changes from JS.

alvarotrigo:
Thanks!

PR #1982, opened 2016-04-24
Fix: scrollToAnchor may not be called

zhangciwu:
on load will not be called if it's binded after onload event,
and jquery ready function will always be called whether document is ready.
(if document has been ready, it will be called immediately)

Found this problem cause I use fullpage in ajax callback (when onload has passed)

### DIFF
--- a/jquery.fullPage.js
+++ b/jquery.fullPage.js
@@ -518,7 +518,7 @@
             //setting the class for the body element
             setBodyClass();
 
-            $window.on('load', function() {
+            $(function(){
                 scrollToAnchor();
             });
         }
--- a/jquery.fullPage.js
+++ b/jquery.fullPage.js
@@ -518,9 +518,10 @@
             //setting the class for the body element
             setBodyClass();
 
-            $(function(){
+            if(document.readyState === 'complete'){
                 scrollToAnchor();
-            });
+            }
+            $window.on('load', scrollToAnchor);
         }
 
         function bindEvents(){


alvarotrigo:
What you suggest is using the same as `$(document).ready(function() {`. Correct me if I'm wrong.

But the problem with that is images might not have been loaded and therefore it can scroll you to the wrong place as slimScroll when using `scrollOverflow` is applied [only on load](https://github.com/alvarotrigo/fullPage.js/blob/master/jquery.fullPage.js#L647), after all images are ready.

zhangciwu:
Yes, `$(function(){...}) `is a shortcut for `$(document).ready(function(){...})`, just shorter.
I've only modified `scrollToAnchor`, 
the `scrollOverflow` you mentioned is still there using onload

zhangciwu:
You're right, 
So now I switch back to onload,
but with a `document.readyState` check to make sure `scrollToAnchor` always be called

alvarotrigo:
Thanks for it!
Can you please provide an example in which the scrollToAnchor would not be called?

zhangciwu:
like this:

`$(function(){
	$.ajax('...some api...')
		.done(function(data){
			...maniplating doms...
			...add extra fullpage sections...

			$('#fullpage').fullpage(...options...)

		})  
})`

alvarotrigo:
So I guess in that scenario you would be destroying fulpage.js and then initializing it again?
And therefore using the scrollToAnchor with the current anchor link of the previous initialization just to scroll again (without animation I suppose) to the active section again?

zhangciwu:
No,
the scenario is that before fullpage called, 
I use css to show the first slide with 100% width and 100% height and a loading image within it,
and hide all other things, user can only seen the first slide and can not scroll cause others are hidden.

Until ajax callback, then add some slides from ajax data, then call fullpage and hide the loading image, 
Now user can scroll to next pages.

So,
First slide fixed on screen  -->   ajax  -->  full page --> scrollable

alvarotrigo:
I see. Makes sense.
Thanks for the explanation!

PR #2003, opened 2016-05-10
using $.filter() in setOptionsFromDOM()

esoniq:
Allows use of any kind of selector as options.sectionSelector


alvarotrigo:
Thanks for it!

PR #2020, opened 2016-05-23
AllowScrolling added to options

Invizible:
I want to have ability to set 'allowScrolling' to false when I create fullpage, so that I can disable scrolling for mouse and keyboard, and make scroll working only with buttons on the page, which will trigger 

```
$.fn.fullpage.setAllowScrolling(true);
$.fn.fullpage.setKeyboardScrolling(true);
```

### DIFF
--- a/jquery.fullPage.js
+++ b/jquery.fullPage.js
@@ -147,6 +147,7 @@
             keyboardScrolling: true,
             animateAnchor: true,
             recordHistory: true,
+            allowScrolling: true,
 
             //design
             controlArrows: true,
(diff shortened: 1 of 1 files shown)


alvarotrigo:
Actually, that's a bug that needs to be solved. 

Doing something like this should work,but it seems it doesn't since fullpage.js 2.7.1 

```javascript
$('#fullpage').fullpage({
    afterRender: function() {
       $.fn.fullpage.setAllowScrolling(false);
    }
});
```

Invizible:
@alvarotrigo okay, I can take a look at this issue and try to fix

PR #2027, opened 2016-05-25
Typo on main file name

bordeo:
My Gulp process with asset-builder can't get the correct file because of typo


alvarotrigo:
Thanks for it!

PR #2029, opened 2016-05-26
fix bug for safari when anchor is unicode 

yyken:
if the value of window.location.hash is Chinese, auto scroll will fail on safari


alvarotrigo:
Thanks for it!

PR #2051, opened 2016-06-02
Fix scrollbar drag issue and mobile touch drag issue on autoScrolling false and fitToSection true.

Vemon:
Symptom: when you turn off `autoScrolling` and turn on `fitToSection`, then try to navigate using the scrollbar, the page doesn't transition smoothly, it always try to fit section to screen and ignore scrollbar dragging, creates some kind of flickering effect. The same sometimes occurs on mobile when user tries to scroll the page up/down.

You can reproduce the issue here on your example page: http://alvarotrigo.com/fullPage/examples/normalScroll.html

Reproduction: Drag the scrollbar so that the section's edge is in the middle of the screen (don't release the scrollbar yet), then as soon as the page starts fitting the section to the screen, start dragging the scrollbar again, you'll see the problem. (It's actually easier to see if you set the fitToSectionDelay to 100).

Side note: I tried to detect the scrollbar mousedown by using `e.target !== $('html').get(0)`, but that doesn't work on Mac, so I'm using mousedown on `$(document)` instead.

### DIFF
--- a/jquery.fullPage.js
+++ b/jquery.fullPage.js
@@ -1700,7 +1700,7 @@
 
         //prevent scrolling on mousedown and/or touchstart
         function interactionStartHandler(e) {
-            if (options.fitToSection) {
+            if (!options.autoScrolling && options.fitToSection) {
                 //if the page is currently scrolling, stop it
                 $htmlBody.stop();
                 //prevent fitToSection while scrollbar is mousedowned
@@ -1710,7 +1710,7 @@
 
         //enable fitToSection on mouseup and/or touchend
         function interactionEndHandler(e) {
-            if (options.fitToSection) {
+            if (!options.autoScrolling && options.fitToSection) {
                 //re-enable fitToSection
                 canScroll = true;
                 //scroll to fit whole section
--- a/jquery.fullPage.js
+++ b/jquery.fullPage.js
@@ -97,7 +97,7 @@
         disableMouse: true,
 
         //fixing bug in iScroll with links: https://github.com/cubiq/iscroll/issues/783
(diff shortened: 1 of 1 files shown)


alvarotrigo:
Thanks for it!

But with your solution the scrolling will stop if the user clicks anywhere in the document (even by mistake), which I believe shouldn't happen.

Vemon:
Yeah, it's a tradeoff, but it's still better than having that flickering effect on the website. Besides, I believe that when user clicks on the website while it is scrolling, they actually want it to stop if there's something interesting bump into their eyes so they can see it. I sometimes do that to take a look at some details in the edge of a section when they're about to go pass the middle of the screen.
Maybe we should add it as an option?

alvarotrigo:
There should be a better approach to it. 
For example:

- Detect mousedown 
- Detect site scrolling while mouse still down (user is dragging the scroll bar)
- Do not fire fitToSection until they release the mouse

Vemon:
Detect mousedown -> detect site scrolling while mouse still down (user is dragging the scroll bar) approach has a flaw. If user simply mousedowns on the scrollbar but not dragging, then fitToSection comes into effect, at that point the position of the page is already different from its previous position which is **remembered** by the scrollbar. Well because fitToSection cannot move the scrollbar which is being hold by the mouse. Then when user starts to drag the scrollbar, flickering will occur, because first the page will jump to the previous position which the scrollbar **remembers**, only then it will start to move to the position user drags the scrollbar to.

alvarotrigo:
Then we need to keep looking for alternatives.

PR #2069, opened 2016-06-06
Fix link to scrolloverflow.min.js

mortonfox:


alvarotrigo:
Thanks for it!
But can you please submit it again using an absolute path instead? 

`https://github.com/alvarotrigo/fullPage.js/vendors/scrolloverflow.min.js`

This way it would still work from other docs such as [npm ones](https://www.npmjs.com/package/fullpage.js).

mortonfox:
Alright, I've changed the patch to use a full path. As you can see though, the path needs a "blob/branch" piece between the repo and the rest of the path.

alvarotrigo:
Thanks for it!

Issue #2075, opened 2016-06-07
Can't reinitialize the plugin after calling destroy() 


rajjoshi22:
### Description
I've been using fullPage.js to build out my own personal site for the last couple days and I ran into an issue where disabling the plugin using destroy() made the plugin completely inactive on a given page until the page was reloaded. 

### Link to isolated reproduction with no external CSS / JS
[www.rajjoshi.site/isolated](url)
Above is a live link to a completely isolated reproduction of the issue on my website (I've never shared code on codepen or jsfiddle yet so this was more convenient). It contains no CSS or JS whatsoever except jQuery version 1.8.1 and non-minified fullPage.js version 2.7.9.

### Steps to reproduce this error on your sites
1. Call ".destroy()" on the fullpage plugin container for whatever reason. 

### Versions 
Tested on the latest versions of Chrome and Microsoft Edge.

alvarotrigo:
reBuild is not the same as "initialise".

> Updates the DOM structure to fit the new window size or its contents. Ideal to use in combination with AJAX calls or external changes in the DOM structure of the site.

Just put your initialization inside a function and call it again.

PR #2076, opened 2016-06-07
Added new function recreate() 

rajjoshi22:
This pull request is in response to the issue I recently posted (issue #2075). 

Essentially, calling `$.fn.fullpage.destroy();` on the container at anytime would disable the plugin on that page permanently until refresh. 

The added function `$.fn.fullpage.recreate()`, when called, re-initializes the plugin on the page without the need to reload the page. 

The plugin is reinitialized to having all the default events handled by the plugin bound. During re-initialization, this function also defaults the values of auto scrolling and keyboard scrolling in all directions to true.


alvarotrigo:
Thanks, but it doesn't make much sense to add an unnecessary function. 
Just placing your initialisation inside a function would do it. 

```javascript
function reCreate(){
   $('#fullpage').fullpage();
}
```

rajjoshi22:
Hey but I _did_ try this. In the live link I provided if you navigated down to the third page I made a third button runs the above code and it didn't work. It was after trying your suggested code and having it not work in the both the latest Chrome and the latest Edge versions that I posted the issue and subsequently added the function.

alvarotrigo:
You need to use `destroy('all')` to destroy the whole. You only destroyed the events.

rajjoshi22:
Thanks, that clears it up. My specific use case was that I still wanted to keep the formatting but didn't need the plugin to be active at specific times.

PR #2078, opened 2016-06-08
Add option for section nav container

pixelmachine:
I've added an option to change the section nav's container so it can appended to elements other than the body (which remains default).

Useful in cases where you're doing things like ajax loading new page fragments that don't use fullPage.js or where you need more control over positioning.


alvarotrigo:
Closed temporarily.

PR #2083, opened 2016-06-09
remove class RESPONSIVE from $body in destroyStructure function

setheid:
This was tripping me up for awhile. After destroying fullPage the fp-responsive class remained on the body and when I re-initialized fullPage it wouldn't go into responsive mode. Removing this class from the body when destroying fullPage fixed this problem.

Great plugin btw! Thank you!

### DIFF
--- a/jquery.fullPage.js
+++ b/jquery.fullPage.js
@@ -97,7 +97,7 @@
         disableMouse: true,
 
         //fixing bug in iScroll with links: https://github.com/cubiq/iscroll/issues/783
-        click: true 
+        click: true
     };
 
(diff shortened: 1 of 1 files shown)


alvarotrigo:
Thanks for it!
Just a little change. We don't need the condition:

Instead of this:

```javascript
if ($body.hasClass(RESPONSIVE)) {
    $body.removeClass(RESPONSIVE);
}
```

Just use this:

```javascript
$body.removeClass(RESPONSIVE);
```

setheid:
Ok, made the change.
Thanks!

alvarotrigo:
Thanks for it!

PR #2203, opened 2016-08-16
removes redundant getTransforms

ironprogrammer:
getFullpageData sets getTransforms option twice, causing issues in IE


alvarotrigo:
Thanks for it @ironprogrammer !

alvarotrigo:
I'll update tomorrow the dist files.

PR #2334, opened 2016-10-17
Bugfix: continuousHorizontal should default to false

publicJorn:
continuousHorizontal was set to `true` resulting in a console warning.
Updated to `continuousHorizontal: false`.

Also made the warning clearer by showing what extension is requested.


alvarotrigo:
Good catch!
Thanks for it!

PR #2373, opened 2016-11-03
Use $ instead of jQuery

GustavSt:
I got error since I didn't have jQuery globally defined. All other places in the code `$` is used as well, not sure why `jQuery` was used here


alvarotrigo:
why did you close it?

PR #2378, opened 2016-11-05
fullPage without jQuery some fixes

qertis:
babel import fix;
firefox scroll fix;
space link fix (url like: /#some thing)


alvarotrigo:
Thanks for that!

PR #2410, opened 2016-11-16
Removing unecessary Prefixes

thisisfed:
http://shouldiprefix.com/#transforms
http://shouldiprefix.com/#transitions

### DIFF
--- a/jquery.fullPage.js
+++ b/jquery.fullPage.js
@@ -2081,7 +2081,6 @@
 
             element.removeClass(NO_TRANSITION);
             return element.css({
-                '-webkit-transition': transition,
                 'transition': transition
             });
         }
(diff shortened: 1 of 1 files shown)


alvarotrigo:
Thanks for it! 
But in this case, I think I would prefer to keep them to maintain older versions just in case. 

It seems `-webkit` still being used in the latest version of *UC Browser for Android* (used by 9% globally [according to caniuse.com](http://caniuse.com/#search=transition), more than IE 10 and 11 combined or than the latest versions of Firefox and Chrome!).

Regarding `moz` and `o`, I guess it doesn't make sense to remove them from a function which purpose is to detect old browsers with no support for translate3d transitions?
The `moz` prefix, for example, [is required in Firefox 10 and Opera 22 to support `translate3d`](http://caniuse.com/#feat=transforms3d). 

Just in case some people is using those old browsers I believe those two lines can be kept in there :)

Issue #2409, opened 2016-11-15
Update docs about media autoplay on iOS


jagi:
I think it's worth adding info about issue with iOS browser and lack of possibility to autoplay media files which is described here: http://stackoverflow.com/a/32945440/1584746
I can create PR with an update if you wish

jagi:
However it looks like it will change in near future http://stackoverflow.com/a/39777974/1584746 but for now it's still not working.

alvarotrigo:
Yeah, feel free to create a PR with a small note about it.
Just a quick mention should be ok. It is not specific for fullPage.js although I guess it can serve as a remainder for some people. 

Maybe something like:

> The autoplay feature might not work on some mobile devices depending on the OS and browsers. 

Not really planning to be updating it for every new OS update or browser version. So a general note should be more than enough.

jagi:
Yes in deed. It's not worth updating and I think it could be even impossible. Apple's mobile Safari is just far behind other browsers and they just started implementing features that were available in other mobile browsers years ago. Ok I wlll create PR tomorrow morning.

alvarotrigo:
Fixed in 2.8.9

PR #2412, opened 2016-11-16
#2409 Add info about lack of autoplay feature on some devices

jagi:


alvarotrigo:
Thanks for it!

jagi:
No problem, I should be thanking you for a great tool :)

PR #2420, opened 2016-11-18
Offset top option

passion4code:
Super simple, and unobtrusive adjustment based on other comments on issues in this project and a professional project in need to allow for fixed headers within a page that has unique full height graphics between each section.

I'm not sure how you go about your minifying and versioning process for pull requests. Open to make adjustments based on feedback. At a glance, I do feel, however, that this small bit is only a value-add and doesn't introduce any issues to your library.

Thanks!


alvarotrigo:
Can you be more specific with the problem it solves?
Can you please add an example of the issue in jsfiddle or codepen?

passion4code:
Padding top places a the padding-top value to the top of each section, which creates a gap between sections that you can see in the scrolling transitions. 

Take a look at this page: https://na.merch.riotgames.com/en/landing/2016/battle-for-baron

That page is using the current unmodified 2.8.8 version

Each section has a unique graphic backdrop that requires seamless transitioning, while also maintaining the need to have a fixed header at the top and not cropping off areas of the section due to the fixed position element.

This offset allows for the fixed position header at the top to more or less feel as though it's a part of the math that determines scrolling positioning and screen height.

I do not have the modified version on a production website at the time, but having the offset will allow for framing design elements to be present on the very top of each section without getting cutoff.

Hopefully that context helps.

passion4code:
Also, happy to add an example file later tonight.

alvarotrigo:
> while also maintaining the need to have a fixed header at the top and not cropping off areas of the section due to the fixed position element.

I guess you are talking about the [2nd section in this case](http://jsfiddle.net/97tbk/1550/)? Where part of the orange background gets hidden behind the header?

passion4code:
Yes, thanks for putting that together real quick. 

Using colors masks the specificity of the scroll points, but it doesn't work as well when each section uses fluid graphics that have defined edges.

passion4code:
Side note - thanks for putting this library together. Saved me probably a week or so of having to build out certain elements of it myself :)

MikeiLL:
I'm hoping this PR can be merged as well. Just discovered this plugin and it looks like it will be a good fit for current project. More so with the offSet option.

PR #2443, opened 2016-12-07
Prevent keypresses to fire prematurely

publicJorn:
Hi Alvaro,

I noticed that the `onLeave` callback was firing too soon, which caused my implementation to break when pressing down (next slide) multiple times. The callback would fire, even though the previous animation was not completed yet.

So I added this check in the `keydownHandler()` function. It's not in the same if-statement as the others, because this would sometimes cause a normal "down" event to fire, scrolling the page one tick, showing a piece of the next section, and then centering on the proper section again.

Hope you find this helpful.

Regards,
Jorn

### DIFF
--- a/jquery.fullPage.js
+++ b/jquery.fullPage.js
@@ -2546,11 +2546,13 @@
         * Adds the possibility to auto scroll through sections on touch devices.
         */
         function addTouchHandler(){
-            if(options.autoScrolling && (isTouchDevice || isTouch)){
+            if(isTouchDevice || isTouch){
                 //Microsoft pointers
                 var MSPointer = getMSPointer();
 
-                $body.off('touchmove ' + MSPointer.move).on('touchmove ' + MSPointer.move, preventBouncing);
+                if(options.autoScrolling){
+                    $body.off('touchmove ' + MSPointer.move).on('touchmove ' + MSPointer.move, preventBouncing);
+                }
 
                 $(WRAPPER_SEL)
                     .off('touchstart ' +  MSPointer.down).on('touchstart ' + MSPointer.down, touchStartHandler)
--- a/jquery.fullPage.js
+++ b/jquery.fullPage.js
@@ -187,7 +187,6 @@
 
         //flag to avoid very fast sliding for landscape sliders
         var slideMoving = false;
-        var sectionMoving = false;
 
         var isTouchDevice = navigator.userAgent.match(/(iPhone|iPod|iPad|Android|playbook|silk|BlackBerry|BB10|Windows Phone|Tizen|Bada|webOS|IEMobile|Opera Mini)/);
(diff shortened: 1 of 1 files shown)


alvarotrigo:
Can you please add a reproduction of the issue?
I don't seem to get any issue on [this example page](http://alvarotrigo.com/fullPage/examples/callbacks.html#secondPage/2).

publicJorn:
See this page: https://www.nrc.nl/nieuws/2016/11/09/ongeloof-bij-hillary-fans-a1530890
What happens?
- On pageload the first element is duplicated and some html added.
- When dom manipulation is ready, fullpage.js will be initialised.
- When you scroll from 1 to 2, the first element is fixed, so you get the effect of a layer being removed. `afterLoad` this element won't be fixed anymore, but since the second image is the same it appears to keep standing.
- When you scroll from 2 to 1, the first element will be fixed again on `onLeave` for the "layer over" effect.

Reproduce:
- Start at the topmost section and tap the down key multiple times successively
- Also, move to down to (ie. the 4th) section and tap the up key multiple times

The `onLeave` function gets fired too soon (this is production, so no console.logs), so the first element doesn't get toggled `fixed` at the right time.

Note: Currently my fix is being put in production, so the error might not be visible for very long.

alvarotrigo:
Please add an isolated reproduction is provided as detailed in the [reporting issues guidelines](https://github.com/alvarotrigo/fullPage.js#reporting-issues) and the [issue skeleton](https://github.com/alvarotrigo/fullPage.js/blob/master/.github/ISSUE_TEMPLATE.md).

publicJorn:
Sorry, should have read the guidelines. Here you go:

# Description
[http://jsfiddle.net/publicJorn/s17s35k5/1/]()

## Steps to reproduce it
- After fiddle loads, click once on the result window to give it focus
- Normal flow:
  - Press down arrow once (fullpage will move to next section)
  - Press down arrow again (fullpage will move to next section)
  - Press up again until you are back at the top
- Errorneus flow:
  - At the top, press down arrow multiple times, quickly after each other
  - The intended effect is not achieved
  - Now, make sure you are at slide nr 3 or 4.
  - Quickly after each other, press up multiple times
  - Also, the wanted effect is not achieved


## Versions
Latest fullpage
Any browser or os

alvarotrigo:
Thanks for the report and the proposed solution!

Although I think I'll go for another approach. 
The proposed solution will prevents sliding horizontally while scrolling. Which is permitted at the moment.
It also creates a duplicate state `sectionMoving` which is the same as the current `canScroll`.

I would go for this:

```javascript
            //do nothing if we can not scroll or we are not using horizotnal key arrows.
            if(!canScroll && [37,39].indexOf(e.which) < 0){
                return;
            }
```

Feel free to change your commit and I'll merge it.

publicJorn:
Neat. Don't know how I missed `canScroll`. Here you go

alvarotrigo:
Thanks for it!

PR #2563, opened 2017-02-14
Patch 1

JavaCowboyArvind:
While reading in scrollOverflow: true, the slide scrolls automatically on reaching end (of section). This can be avoided by adding a minimum pause of 150 before scrolling to next slide


alvarotrigo:
Thanks for that! But unfortunately that's not a good solution.
Kinetic scrolling users might get stuck on the bottom if they keep scrolling and the inertial scrolling keeps getting catch. 

Also, it will prevent to scroll up or down in sections with no scrolloverflow.

PR #2725, opened 2017-05-25
fix #2706 back history

### LINKED ISSUE #2706: Back history to 1st horizontal slide
### Description
When i go back from URL like  #section/slide2 to #section/slide1 it work perfect, and function  scrollPageAndSlide work good. But when i go back to #section it not go to slide 0.

When i debug, i see this code
```
function hashChangeHandler(){
            if(!isScrolling && !options.lockAnchors){
                var value =  window.location.hash.replace('#', '').split('/');
                var sectionAnchor = decodeURIComponent(value[0]);
                var slideAnchor = decodeURIComponent(value[1]);
....
                        scrollPageAndSlide(sectionAnchor, slideAnchor);
....
}
```
decodeURIComponent return string 'undefined' when not set.

```
function scrollPageAndSlide(destiny, slide){
            var section = getSectionByAnchor(destiny);

            //do nothing if there's no section with the given anchor name
            if(!section.length) return;

            //default slide
            if (typeof slide === 'undefined') {
                slide = 0; **//not comes here**
            }
....
}
```
I assume that is not correct to compare a string 'undefined' with the type of undefined ?

### Versions 
[ * fullPage 2.9.4, Firefox 53.0.2 and posible any others]

talgautb:
```
typeof slide === 'undefined'
```
is not working, because it is `string`.
maybe some other check for `undefined` ?

### DIFF
--- a/jquery.fullPage.js
+++ b/jquery.fullPage.js
@@ -1752,6 +1752,9 @@
                     if ((sectionAnchor && sectionAnchor !== lastScrolledDestiny) && !isFirstSlideMove || isFirstScrollMove || (!slideMoving && lastScrolledSlide != slideAnchor ))  {
                         scrollPageAndSlide(sectionAnchor, slideAnchor);
                     }
+                } else {
+                    // go to first slide
+                    moveTo(1);
                 }
             }
         }
@@ -2327,7 +2330,7 @@
             if(!section.length) return;
 
             //default slide
-            if (typeof slide === 'undefined') {
+            if (slide === 'undefined') {
                 slide = 0;
             }
 


alvarotrigo:
Thanks for that! But I don't think I understood well what problem you are trying to solve here. 
Any particular one that stills one in the issues tracker? Which ? If not, what's about?

Regarding the condition, `slide` can indeed be undefined. Once can do `$.fn.fullpage.moveTo(3)`, which will cause `slide` to be `undefined` within `scrollPageAndSlide`, so the check is ok in that case.
For which case scenario you think it is wrong?

talgautb:
it is about #2706 it is not show first slide.
`slide to be undefined within scrollPageAndSlide, so the check is ok in that case`
my fault, ok

PR #2728, opened 2017-05-27
Proper viewport size, do not update height in responsive mode, do not forget to rebuild on touch devices if change is less than 20%

tmcdos:
I would like to share a contribution. I am using the library for the landing page of a customer's web site - development copy is currently available at [http://mmx.ivogelov.com](http://mmx.ivogelov.com)
I have experienced some strange/erroneous behavior in a couple of cases:

- **fullPage.js** ignores the scrollbar in its initialization - this means that the computed height for the sections is less than the full viewport height by the height of the horizontal scrollbar. To solve the issue, I am using a cross-browser compatible function to find the proper viewport height. This open-source function is [viewportSize](https://github.com/tysonmatanich/viewportSize) - I replaced all occurences of **$window.height()** with **viewportSize.getHeight()**
- **fullPage.js** tries to apply section heights in **resizeHandler()** even if the **responsive** mode is currently active - this breaks the page layout because in this mode sections' height should be set to AUTO and not stretched to the viewport height. To solve the issue, I immediately exit from the **resizeHandler()** if BODY has a CSS classname of **fp-responsive**
- **fullPage.js** *forgets* to rebuild the sections on touch devices if the change in height is less than 20%. To solve the issue, I start a timer - if there is no new change before the timeout, a rebuild is forced after the timeout elapses

### DIFF
--- a/jquery.fullPage.js
+++ b/jquery.fullPage.js
@@ -1,3 +1,6 @@
+/*! viewportSize | Author: Tyson Matanich, 2013 | License: MIT (https://github.com/tysonmatanich/viewportSize) */
+(function(n){n.viewportSize={},n.viewportSize.getHeight=function(){return t("Height")},n.viewportSize.getWidth=function(){return t("Width")};var t=function(t){var f,o=t.toLowerCase(),e=n.document,i=e.documentElement,r,u;return n["inner"+t]===undefined?f=i["client"+t]:n["inner"+t]!=i["client"+t]?(r=e.createElement("body"),r.id="vpw-test-b",r.style.cssText="overflow:scroll",u=e.createElement("div"),u.id="vpw-test-d",u.style.cssText="position:absolute;top:-1000px",u.innerHTML="<style>@media("+o+":"+i["client"+t]+"px){body#vpw-test-b div#vpw-test-d{"+o+":7px!important}}<\/style>",r.appendChild(u),i.insertBefore(r,e.head),f=u["offset"+t]==7?i["client"+t]:n["inner"+t],i.removeChild(r)):f=n["inner"+t],f}})(this);
+
 /*!
  * fullPage 2.9.4
  * https://github.com/alvarotrigo/fullPage.js
@@ -198,7 +201,7 @@
         var isTouchDevice = navigator.userAgent.match(/(iPhone|iPod|iPad|Android|playbook|silk|BlackBerry|BB10|Windows Phone|Tizen|Bada|webOS|IEMobile|Opera Mini)/);
         var isTouch = (('ontouchstart' in window) || (navigator.msMaxTouchPoints > 0) || (navigator.maxTouchPoints));
         var container = $(this);
-        var windowsHeight = $window.height();
+        var windowsHeight = viewportSize.getHeight(); // IVO GELOV = $window.height();
         var isResizing = false;
         var isWindowFocused = true;
         var lastScrolledDestiny;
@@ -450,7 +453,7 @@
 
             isResizing = true;
 
-            windowsHeight = $window.height();  //updating global var
+            windowsHeight = viewportSize.getHeight(); // IVO GELOV = $window.height();  //updating global var
 
             $(SECTION_SEL).each(function(){
                 var slidesWrap = $(this).find(SLIDES_WRAPPER_SEL);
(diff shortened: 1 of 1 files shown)


jquery.fullPage.js
```diff
@@ -1,3 +1,6 @@
+/*! viewportSize | Author: Tyson Matanich, 2013 | License: MIT (https://github.com/tysonmatanich/viewportSize) */
+(function(n){n.viewportSize={},n.viewportSize.getHeight=function(){return t("Height")},n.viewportSize.getWidth=function(){return t("Width")};var t=function(t){var f,o=t.toLowerCase(),e=n.document,i=e.documentElement,r,u;return n["inner"+t]===undefined?f=i["client"+t]:n["inner"+t]!=i["client"+t]?(r=e.createElement("body"),r.id="vpw-test-b",r.style.cssText="overflow:scroll",u=e.createElement("div"),u.id="vpw-test-d",u.style.cssText="position:absolute;top:-1000px",u.innerHTML="<style>@media("+o+":"+i["client"+t]+"px){body#vpw-test-b div#vpw-test-d{"+o+":7px!important}}<\/style>",r.appendChild(u),i.insertBefore(r,e.head),f=u["offset"+t]==7?i["client"+t]:n["inner"+t],i.removeChild(r)):f=n["inner"+t],f}})(this);
```
tmcdos: This is a cross-browser safe way to get the proper viewport size (does not include scrollbars) - because **$window.height()** returns a value which is the viewport height minus the height of the horizontal scrollbar.

jquery.fullPage.js
```diff
@@ -2076,22 +2080,33 @@
         function resizeHandler(){
             //checking if it needs to get responsive
             responsive();
+            if(document.body.classList.contains('fp-responsive')) return; // IVO GELOV
```
tmcdos: When the **responsive** mode is active, we should not enforce the viewport height to the sections - their height should be left to AUTO.

jquery.fullPage.js
```diff
@@ -2076,22 +2080,33 @@
...
-                    var currentHeight = $window.height();
+                    var currentHeight = viewportSize.getHeight(); // IVO GELOV = $window.height();
 
                     //making sure the change in the viewport size is enough to force a rebuild. (20 % of the window to avoid problems when hidding scroll bars)
-                    if( Math.abs(currentHeight - previousHeight) > (20 * Math.max(previousHeight, currentHeight) / 100) ){
-                        reBuild(true);
-                        previousHeight = currentHeight;
+                    if( Math.abs(currentHeight - previousHeight) > (20 * Math.max(previousHeight, currentHeight) / 100) ) reBuild(true); // IVO GELOV
+                    else
+                    {
```
tmcdos: If the change is less than 20%, we should not ignore the event - just postpone the rebuild for some time,

jquery.fullPage.js
```diff
@@ -2076,22 +2080,33 @@
...
+                      // IVO GELOV =====
+                      clearTimeout(resizeId);
+      
+                      resizeId = setTimeout(function(){
+                          reBuild(true);
+                      }, 350);
+                      // IVO GELOV =====
                     }
+                    previousHeight = currentHeight; // IVO GELOV
```
tmcdos: I am not sure whether the above statement should be inside the ELSE branch or outside of it ....

alvarotrigo:
Thanks for that!
I'll review it later on whenever I have more time, but so far:

> fullPage.js tries to apply section heights in resizeHandler() even if the responsive mode is currently active - this breaks the page layout because in this mode sections' height should be set to AUTO and not stretched to the viewport height. To solve the issue, I immediately exit from the resizeHandler() if BODY has a CSS classname of fp-responsive

fullPage.js should set the height of the sections even in responsive mode. 
Use the class `fp-auto-height-responsive` if you want to avoid that behaviour.

> fullPage.js forgets to rebuild the sections on touch devices if the change in height is less than 20%. To solve the issue, I start a timer - if there is no new change before the timeout, a rebuild is forced after the timeout elapses

That prevents fullpage.js to resize the section when the address and bottom bars get hidden. When using scrollBar:true or autoScrolling:false, this might be an issue. 
Have you tested those cases?

tmcdos:
> When using scrollBar:true or autoScrolling:false, this might be an issue.

I am using the default values for all parameters - I only define the **responsiveWidth** and provide handlers for the **afterResize** and **afterResponsive** events when initializing the library. I have not tested with `scrollBar:true` or `autoScrolling:false`.

> Use the class `fp-auto-height-responsive` if you want to avoid that behaviour.

You are right - using this class solved the issue and did not require the presense of my code for early exiting the **resizeHandler()** in responsive mode. I will revert the changes.

PR #2790, opened 2017-07-09
fix destroy method

qaz3954179:
Remove jquery.fn.data's cache for anchor when destroy.

### DIFF
--- a/jquery.fullPage.js
+++ b/jquery.fullPage.js
@@ -2779,6 +2779,10 @@
                 setSrc($(this), 'srcset');
             });
 
+            container.find('[data-anchor]').each(function () {
+                $(this).removeData('anchor');
+            });
+
             $(SECTION_NAV_SEL + ', ' + SLIDES_NAV_SEL +  ', ' + SLIDES_ARROW_SEL).remove();
 
             //removing inline styles


alvarotrigo:
Thanks for it!
Could you please explain the problem it fixes? And where in fullPage.js is used the `data` method to set the anchor value?

qaz3954179:
In the same page, no reload.
Firstly, I invoke it, eg: "$.fullpage({anchor:[anchor_1, anchor_2]})"
Secondly, to destroy it, and reinvoke it with different anchor。 eg: $.fullpage({anchor: [anchor_sec_1, anchor_sec_2]})
when i scroll to the top, the window.location.hash's value is the first one. eg: url/#anchor_1.

alvarotrigo:
Right I see now! Thanks for the explanation!
Although you won't need to use `removeData`. Removing the attribute should do it. `removeAttr`.A

And it should only do it when the anchors are defined in the fullpage.js initialization object, not when they are defined in the DOM with the `data-anchor` attribute. So we'll need a condition there or an extra flag to figure out if it was defined in one or another way.

qaz3954179:
Ah, I'm not sure, in that case, when they are not defined.
But, we can't use the `removeAttr `, we should to clean the cache with `removeData`, not remove the attribute.
Actually, the Dom' attribute is the new value, but, the `$.fn.data` method don't get it.

alvarotrigo:
> But, we can't use the removeAttr, we should to clean the cache with removeData, not remove the attribute.

Why.
The attribute should be removed if it was added by fullPage.js. 
And it shouldn't when it wasn't. (defined by the user in the DOM)

qaz3954179:
oh, you are right.
I think, this is another question.
The solution to that question is like what you said.

>  need to use removeData 
>  So we'll need a condition there or an extra flag to figure out if it was defined in one or another way.
i had re-push it.

alvarotrigo:
You defined the flag to `true` by default. It shouldn't be like that.
You should probably look at it here [in this line](https://github.com/alvarotrigo/fullPage.js/blob/master/jquery.fullPage.js#L627).

Let me know if you still want to commit the pull request to `dev`or not.

Then add it in the previous loop:

```javascript
container.find('[data-anchor]').each(function () {          
     $(this).removeData('anchor');
     if( anchorsDefinedByUser){
           //do whatever
     }
});
```

PR #2903, opened 2017-09-13
Allow forced scroll to top or bottom of section.

avoliva:
I was using this library but I had a need to always force the section to align to the bottom instead of changing depending on where you're scrolling from. Setting `forcedPosition` to `bottom` has the section always position at the bottom of the viewport, `top` the opposite. By default it's `null` which allows it to function as it currently does.


alvarotrigo:
Thanks for it, I'll review it at some point! Although you should commit it to the `dev` branch in any case!

avoliva:
Thanks I completely forgot about committing to the dev branch

PR #2907, opened 2017-09-19
Allow forced scroll to top or bottom of section.

avoliva:
I was using this library but I had a need to always force the section to align to the bottom instead of changing depending on where you're scrolling from. Setting `forcedPosition` to `bottom` has the section always position at the bottom of the viewport, `top` the opposite. By default it's `null` which allows it to functional as it currently does.

Same as #2903, just comitted to dev branch

### DIFF
--- a/jquery.fullPage.js
+++ b/jquery.fullPage.js
@@ -150,6 +150,7 @@
             touchSensitivity: 5,
             normalScrollElementTouchThreshold: 5,
             bigSectionsDestination: null,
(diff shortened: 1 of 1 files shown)


alvarotrigo:
Thanks I'll review it at some point!

avoliva:
Was there an issue?

alvarotrigo:
> Was there an issue?

No. But needed to close the issue temporary to create a pull request for the next release.

PR #2912, opened 2017-09-22
Added ability to setup iScroll events.

MaxxSoftware:


alvarotrigo:
Can you please provide a description of the issue you solved and an example of its use?

MaxxSoftware:
You can add a listeners to the iScroll's events the same way as you setup the iScroll's settings by setting scrollOverflowOptions
For example 

`$('#fullpage').fullpage({
  scrollOverflow: true,
  scrollOverflowOptions: [mouseWheelSpeed: 30],
  scrollOverflowEvents: [
    scrollStart: function() { // when iScroll event "scrolStart" is fired.
      if (this->y == 0) {
        console.log('we reach the bottom of the slide with scroll.');
      }
    } 
  ]
});`

alvarotrigo:
Thanks for that! I'll review it whenever I have some time!

TylerOlthuizen:
@MaxxSoftware Hey thanks for the code! I am trying to add this project but Im getting an Uncaught SyntaxError: Unexpected token : Error. I am just using the code you have written and I updated the full page js file. Any ideas why i'm getting this?

alvarotrigo:
@MaxxSoftware I just noticed those events will get fired for every single section / slide that uses scrolloverflow.

There's no way to determine in this way which event belongs to which section without using fullpage.js state classes.

Any suggestions ?

PR #2956, opened 2017-10-24
Add jsDelivr hits badge

LukasDrgon:
Hey, I noticed that fullPage.js gets 632k hits per month on jsDelivr, so I thought you might like this badge.


alvarotrigo:
Awesome!! Thanks for that!!

PR #2989, opened 2017-11-19
Improved: adding color to vertical and horizontal navigation dots

vladcmanea:
This change adds colors to the vertical and horizontal dots. Such a feature is necessary if the color of one of the sections or slides that overlaps with the dots does not really work with the default color. For example, the current default dark gray may not look well on a dark background.

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,6 +156,7 @@ $(document).ready(function() {
 		navigation: false,
 		navigationPosition: 'right',
 		navigationTooltips: ['firstSlide', 'secondSlide'],
+		navigationColor: '#333',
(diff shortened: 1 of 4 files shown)


alvarotrigo:
Thanks for that! Although I believe it is better to leave to CSS all things regarding styling the site. 
Using something like this should be the way to go to change the nav color:

```css
#fp-nav ul li a span, 
.fp-slidesNav ul li a span{
    background: #fff;
}
```

This also give you the freedom to change the color depending on the section/slide using css3 conditions based on the fullpage.js state classes.

PR #3014, opened 2017-11-30
UPDATE?

thisisfed:
Removed some 0 (zeros) in the transform timings 
and added some missing prefixes


alvarotrigo:
Thanks for that!! But I wouldn't worry about that too much :) That's de developing version, the one I use to code. 

You have them all removed [in the minified file](https://github.com/alvarotrigo/fullPage.js/blob/master/dist/jquery.fullpage.min.css) which is the one you should be using for production environments.

PR #3053, opened 2018-01-04
Tell people to load the scrollOverflow script after jQuery

vityavv:
I had a long-standing issue where `scrollOverflow: true` was not doing anything. After a lot of time debugging, I eventually found that loading it after jQuery solved the issue. I think that it is very important for other people to know this so that they don't have to go through the pointless journey I went through.

I feel really stupid now


alvarotrigo:
Apologies for the confusion and thanks for the PR!! ;)

alvarotrigo:
Merged!

PR #3057, opened 2018-01-05
new option: touchDeviceArrows

vityavv:
This is a simple option that hides the device arrows if the user is a mobile phone as defined by `isTouchDevice`

### DIFF
--- a/README.md
+++ b/README.md
@@ -394,6 +394,8 @@ the fitting by the configured milliseconds.
 
 - `keyboardScrolling`: (default `true`) Defines if the content can be navigated using the keyboard.
 
+- `touchDeviceArrows`: (default `true`) Defines if the control arrows should be shown for touchscreen devices like mobile phones
+
 - `touchSensitivity`: (default `5`) Defines a percentage of the browsers window width/height, and how far a swipe must measure for navigating to the next section / slide
 
 - `continuousVertical`: (default `false`) Defines whether scrolling down in the last section or should scroll down to the first one and if scrolling up in the first section should scroll up to the last one. Not compatible with `loopTop`, `loopBottom` or any scroll bar present in the site (`scrollBar:true` or `autoScrolling:false`).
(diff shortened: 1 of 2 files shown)


alvarotrigo:
Thanks for it!!

Although in this case I believe the way to go for is is making use of CSS conditions based on the fullPage.js state classes or callbacks. 
Therefore I'll close the PR.

vityavv:
@alvarotrigo While the css is easy, you already have a `isTouchDevice` variable and so doing has several advantages:
 - The user has to write less code
 - Since the `isTouchDevice` variable is used in other parts of the code it is more likely to be updated, so when user agent strings change (like with new devices) websites who use the option rather than writing their own code will be updated as well (provided they update the version of fullpage.js)

alvarotrigo:
I appreciate your explanation but I would say it doesn't make much sense to think in terms of "Mobile" as "Touch devices". 
Nowadays there's no such thing as a "Mobile". Is a tablet a mobile? Is a computer with touch screen a mobile? Would you keep the arrows in a desktop computer that resizes the viewport to a smaller size than a mobile or a tablet?

The way to deal with a responsive sites is based on the viewport dimensions and not on the touch feature. That's how it is always done. 
For this cases I recommend making use of the responsive options provided by fullPage.js if they want to turn off the autoScrolling. 
If not, they can use the callback `afterResize` and use a condition to show or hide the arrows.

As easy as:

```javascript
afterResize: function(){
     if($(window).width() < 800 && ! arrowsHidden){
          //hide arrows
          arrowsHidden = true;
     }else{
          //show arrows
          arrowsHidden = false;
     }
}
```

Moreover, adding new options to fullPage.js is something I try to avoid unless required. There are already plenty of options and adding options to deal with CSS matters is kind of irrelevant. 
Those things should be directly treated with CSS, where you can do many more things, such as change their color, their shape, their position etc. 
Providing options for all of those cases would be overcomplicating things.

vityavv:
The reason that I checked for `isTouchDevice` is because you can swipe through slides on mobile devices. The reason that I didn't check for `isTouch` is because people on laptops with touchscreens might want to navigate with the keyboard.

PR #3084, opened 2018-01-30
Move getPaddings to upper scope

Odrin:
Move getPaddings func to upper scope since it is not available in iscrollHandler object:
https://github.com/alvarotrigo/fullPage.js/blob/c0019fbc33ee54c82a593d91fa556b572352aa32/vendors/scrolloverflow.js#L2736


alvarotrigo:
Uh! What a mistake! 😄 
Thanks for that!!

PR #3085, opened 2018-01-30
Cancelling the scroll before it takes place

joelangford:
Modified so https://github.com/alvarotrigo/fullPage.js#cancelling-the-scroll-before-it-takes-place works on the pure js version.


alvarotrigo:
Thanks for that!

PR #3114, opened 2018-02-20
Updating gulp-uglify to latest version as well as changing deprecated…

josephahern:
… "preserveComments" with new syntax. This solves the problem of minified versions of fullpagejs from being compatible with webpack bundling.


alvarotrigo:
Thanks a lot for that @josephahern !!

Issue #1237, opened 2015-05-21
"Tab" key = bug


meneldil:
Hello,
I discovered a bug.

If the website contains a link or a form when you press the "tab" key, the browser goes directly to the section concerned but the navigation is completely broken.

alvarotrigo:
Yeah right. 
I was aware of it. 

The only solution is not simple... [Depending on the browser](http://stackoverflow.com/a/1600194/1081396) some elements are able to receive the focus on `tab` and others don't.

In order to prevent it, fullPage.js would have to set `tabIndex="-1"` to any of those elements outside the current section plus any element containing the attribute `tabindex`.

The other option is just to disable the tab in fullPage.js. Or you manually setting tabindex to `-1`.

Criten:
A quick fix we implemented was this...

``` js
  $('a, input, select').prop('tabIndex', -1);
```

This basically does that alarotrigo said and disables tabbing globally on the page... We actually just got a complaint from a client saying that they couldn't tab through this one form so you might be getting a pull request if I can find a decent solution, but as you said it might affect performance if there are a ton of elements...

alvarotrigo:
You can also set the wrapper of those elements to `display:none` or I believe you can even use `opacity:0` and probably...  `visibility: hidden`. This way you will only have to show the form `onLeave` or just use the active class in the section to make it visible:

```css
#section3 #myForm{
    display:none;
}
#section3.active #myForm{
    display: block;
}
```

alvarotrigo:
I'll keep looking into this issue. 
It might be able to be solved by using `opacity:0` for all the non visible sections. I believe that's the way [reveal.js](http://lab.hakim.se/reveal-js) deals with it.
Although they have a pretty ugly internval [here](https://github.com/hakimel/reveal.js/blob/master/js/reveal.js#L582) to solve problems with iframes. (you can actually test it by setting `opacity:1` to the next slide, you'll notice how it comes back)

Criten:
I just did a test with the following.

```js
$('.fp-section').not('.active').css('opacity', 0);
```
and I can still tab to the opaque elements, you might need to use display: none...

alvarotrigo:
True.

Criten:
Seems to work pretty well doing the following..

```js
// On load
$('.fp-section').not('.active').hide();

// On transition - To next slide
$('.fp-section.active').next().show();
```

[...]

PR #3124, opened 2018-02-27
Remove keydown handler on destroy

archferns:
This is made necessary as the fix for  #1237, by adding the `onTab` function is not reverted on destroying fullpage.js plugin. Consequently, this was breaking tab functionality when the plugin is no longer active.

### DIFF
--- a/jquery.fullPage.js
+++ b/jquery.fullPage.js
@@ -2724,6 +2724,7 @@
                 .off('resize', resizeHandler);
 
             $document
+                .off('keydown', keydownHandler)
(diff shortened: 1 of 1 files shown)


alvarotrigo:
Thanks for that! 
Also, we should also unbind the `keyup` event!

archferns:
Of course! I will update the PR.

alvarotrigo:
@archferns just create a new one. I've merged this one :)

alvarotrigo:
Also, if you could change the version on the top, that'd be great! Just add 2.9.7

archferns:
Will do that. thanks!

PR #3125, opened 2018-02-27
Remove keyup handler on destroy

archferns:
1. Complementing #3124, unbinding the keyup handler as well to revert to the original state on destroying fullpage.js.
2. Version bump to 2.9.7


alvarotrigo:
Awesome! Thanks for that!

archferns:
You are welcome. Great library by the way!

alvarotrigo:
Thanks!! ;)

Issue #3170, opened 2018-03-28
CSP - Inline style violates the style-src Content Security Policy directive.


flea89:
### Description
Error: `Refused to apply inline style because it violates the following Content Security Policy directive: "style-src 'self'". Either the 'unsafe-inline' keyword, a hash ('sha256-0LxRaYHmC6CoywhcJZ5U9VA1MWk4/yzUeQtq9QXZeh0='), or a nonce ('nonce-...') is required to enable inline execution.`

### Link to isolated reproduction with no external CSS / JS
https://codepen.io/flea89/full/xWpBEy

### Steps to reproduce it
1. Add `<meta http-equiv="Content-Security-Policy" content="style-src 'self">` in the head
2. Check the console for CSP violations.

alvarotrigo:
Probably a codepen issue and not a fullpage.js one. 
fullPage.js doesn't inline any CSS. Only when using certain extensions.

flea89:
This is not a codepen issue, the problem is caused by  https://github.com/alvarotrigo/fullPage.js/blob/eed567efacf773194078f7d6e051176796f4c6df/jquery.fullPage.js#L2257.

I've created the codepen to show you the issue , and set `<meta http-equiv="Content-Security-Policy" content="style-src 'self">` in the head in order to break it.

The height shouldn't be set through a the `style=` attribute but should be set using `element.style.height` or alternative solutions jquery `.height()`.

alvarotrigo:
> The height shouldn't be set through a the style= attribute but should be set using element.style.height or alternative solutions jquery .height().

Right. I had no idea of such rule :)
If you think this can be solved by applying your suggestions, please feel free to create a pull request on the dev branch. 

Otherwise I'll take a look at that whenever I find some time.

alvarotrigo:
Solved in the new release 2.9.7!

PR #3172, opened 2018-03-29
Use jquery height to set wrapper height, instead on relying on style …

flea89:
Using inline styles violates style-src CSP rule (unless you set `unsafe-inline`).
Set the wrapper height using `.height(value)` instead of setting it using `style` attribute.
#3170


alvarotrigo:
Thanks for that!

PR #3195, opened 2018-04-12
Fix #2865

### LINKED ISSUE #2865: data-srcset in picture source tags doesn't work
### Description
Using data-srcset in the source tags of picture tags does not work.

### Link to isolated reproduction with no external CSS / JS
[Ideally in jsfiddle.net or codepen.io, links to personal websites won't be reviewed unless isolated. Reported issues without a reproduction might get closed.]
https://jsfiddle.net/ibrokemypie/xmbguzt3/7/

### Steps to reproduce it
Create a picture tag
Define sources with data-srcset values
Make the img tag's data-src value incorrect to force browser to load the source tags.
No image shows.
Removing the data- prefix in the source tags allows this to work as expected.

### Versions 
Tested on Chrome dev, firefox nightly and safari on macOS sierra.

prplx:
Fix lazyLoad function to work with picture tag.

### DIFF
--- a/jquery.fullPage.js
+++ b/jquery.fullPage.js
@@ -1624,14 +1624,9 @@
                     }
                 });
 
-                if(element.is('source')){
-                    var types = ['video', 'audio', 'picture'];
-                    var elemParent = element.parent();
-                    var typeToPlay = elemParent.get(0).tagName.toLowerCase();
-
-                    if (types.indexOf(typeToPlay) !== -1) {
-                        elemParent.load();
-                    }
+                if(element.is('source') && !element.closest('picture').length) {
+                    var typeToPlay = element.closest('video').length ? 'video' : 'audio';
+                    element.closest(typeToPlay).get(0).load();
                 }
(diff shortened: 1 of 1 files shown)


alvarotrigo:
Thanks a lot!!

Interesting.... had no idea the picture tag admits a `.load` function! 
Could you please provide an example of this fixture in jsfiddle or codepen?

alvarotrigo:
Could you please provide an isolated reproduction of the issue? 
A picture element that doesn't lazy load properly?

prplx:
> A picture element that doesn't lazy load properly?

Exactly! I made a fiddle https://jsfiddle.net/w9jzyes4/6/. Please open the dev tools console and try to slide to the third section. That's because of the source tag inside the picture. You suppose that there are only two tags with a source tag inside: video and audio.

alvarotrigo:
Thanks for that!!

prplx:
You're welcome!

PR #3217, opened 2018-04-27
Add scrollOverflowOptions prop `firstStepDeceleration`

jnotnull:

1. USAGE:

```
$('#fullpage').fullpage({
    scrollOverflow: true,
    scrollOverflowOptions: {
        firstStepDeceleration: true
        deceleration: 0.0024,
    }
});
```

2. WHY

If i use `scrollOverflow` property, some pages maybe overflow if the heigh of this page is greater than the height of view, but other pages maybe not overflow.  if i slide from page which is not overflow to  page which is over flow, the speed of roll is too heigh. if i use `firstStepDeceleration` property, it may be more smoothly.


alvarotrigo:
Thanks for that!
Could you please provide a jsfiddle example and the steps to reproduce it in ordrr to illustrate the issue you are attempting to solve?

> if i slide from section which is not overflow to section which is over flow, the speed of first roll is too fast at overflow section

I can't seem to reproduce it here:
https://jsfiddle.net/1kyuL4em/44/

Are you talking about sliding with the mouse wheel and trackpad? Or about touch devices?
I'm using a trackpad and everything seems alright to me.

jnotnull:
Ok, thanks for your example, your example is enough for me.
If i am a coder, i know first section is fullpage, the second section is overflow, i maybe sliding my finger unforcely.
But if i am a user, i dont know the second section is overflow, i just know sliding my finger hard because human's thinking is linear.  When i sliding hard when i at second section, this section is scroll sometimes i  cant see the content clearly.

alvarotrigo:
Are you talking about sliding with the mouse wheel and trackpad? Or about touch devices?

jnotnull:
It is touch device, just iphone 8 plus.

alvarotrigo:
I was playing a bit with this but I was not able to notice any difference.
I tried using both, kinetic scrolling trackpad (Macbook pro) and and iPhone 11.

PR #3233, opened 2018-05-14
Make README gender neutral

lauradegroot:
The README specifies that the user would always be male, which is not the case


alvarotrigo:
Thanks for that! ;)

PR #3243, opened 2018-05-23
Kyl3thegreat patch 1

kyl3thegreat:

Fixes issue when looping through the vendor prefixes at the beginning of the code and changes 'ransform', should be, transform. This fixed an issue on mobile devices where I wasn't able to scroll past a certain point on the last section on landscape view, as well as when changing back to portrait from landscape fixes height layouts for each section


alvarotrigo:
I'm not quite sure to understand the fix.
`vendors[i]` is `t` or `webkitT` etc.
So, it ends but being `t` + `ransform`, which is correct?

```javascript
var _vendor = (function () {
        var vendors = ['t', 'webkitT', 'MozT', 'msT', 'OT'], 
            transform,
            i = 0,
            l = vendors.length;

        for ( ; i < l; i++ ) {
            transform = vendors[i] + 'ransform';
            if ( transform in _elementStyle ) return vendors[i].substr(0, vendors[i].length-1);
        }

        return false;
    })();
```

kyl3thegreat:
Ahhh your right, I see now. My mistake

PR #3316, opened 2018-07-17
Prevent jQuery adapter from loading when fullpage is not global

steriley:
If you import fullpage.js into a module without a jQuery dependency and bundled with webpack.

When using the bundle in a page that has jQuery as a global, it will expect fullpage to exist as a global which can cause an error.

By adding window.jQuery && window.fullpage to the conditional it will prevent the adapter from loading unless fullpage is a global.

### DIFF
--- a/src/fullpage.js
+++ b/src/fullpage.js
@@ -3845,7 +3845,7 @@
 /**
  * jQuery adapter for fullPage.js 3.0.0
  */
-if(window.jQuery){
+if(window.jQuery && window.fullpage){
     (function ($, fullpage) {
         'use strict';
 


alvarotrigo:
Hey @steriley thanks a lot for that!! I appreciate the improvement!

Although, not quite sure if I understood the issue properly.
You said:

> When using the bundle in a page that has jQuery as a global, it will expect fullpage to exist as a global which can cause an error.

What error?

Is it because a webpack module won't define fullpage.js in the window object? As it will use [this line](https://github.com/alvarotrigo/fullPage.js/blob/master/src/fullpage.js#L20) instead of [this other one](https://github.com/alvarotrigo/fullPage.js/blob/master/src/fullpage.js#L23)?

```javascript
//when using webpack
module.exports = factory(window, document);

//when not using it
window.fullpage = factory(window, document);
``` 

Now my question is, how would they then be able to use the jQuery adapter in case of using webpack if we apply your suggested improvement?

steriley:
The error that has triggered the pull request is this:

```
    ReferenceError: fullpage is not defined
    at eval (fullpage.js?16fe:3864)
    at Object../node_modules/fullpage.js/dist/fullpage.js (vendors~page-content.js:43)
    at __webpack_require__ (app.js:765)
    ...
```

Our webpack bundle is being loaded into a third party site that we do not have control over and already loads jQuery into the window object (we have no control over when this happens either).

As you have pointed out fullpage is within a webpack module, and not in the window object so the ReferenceError is triggered when the adapter attempts to initialise.

My assumption is that jQuery & fullpage both need to be in the global window object for the adapter to function and that there is little benefit to using the adapter in a webpack setup.

alvarotrigo:
> My assumption is that jQuery & fullpage both need to be in the global window object for the adapter to function and that there is little benefit to using the adapter in a webpack setup.

Yeah I see, so basically if they want to make use of jQuery for fullPage.js, they would have to define fullpage.js in the window object themselves. 

Ok! Got it!

alvarotrigo:
Thanks for that!

PR #3319, opened 2018-07-18
IScroll fix

AntonyMayer:
## Issue

At current project we are using **fullpage.js** npm along with **browserify** as a JS module bundler.
We faced an issue with **IScroll** scope.

Module responsible for initiating **fullpage.js** supposebly should contain:
```javascript
import 'fullpage.js/vendors/scrolloverflow'; // to allow scrollOverflow
import fullpage from 'fullpage.js';

new fullpage('#someID', {
    licenseKey: 'key'
    scrollOverflow: true,
});
```

Here's an error we got thrown from line 2171 of `./vendors/scrolloverflow.js`
```javascript
Uncaught ReferenceError: IScroll is not defined
```

## Solution
The problem is at line 2116 of `./vendors/scrolloverflow.js`:
```javascript
if ( typeof module != 'undefined' && module.exports ) {
    module.exports = IScroll;
} else if ( typeof define == 'function' && define.amd ) {
        define( function () { return IScroll; } );
} else {
    window.IScroll = IScroll;
}
```

**Browserify** is using CommonJS modules definition, hence **IScroll** gets **exported** by being assigned to **module.exports** instead of being exposed to global scope. 

**Scrolloverflow 2.0.0 module for fullPage.js** in `./vendors/scrolloverflow.js` that follows immediately after **IScroll** being exported assumes it to be available in global space and that's causing an error.

I added a check in **Scrolloverflow 2.0.0 module for fullPage.js** for **IScroll** availability in global scope, otherwise it creates a scoped variable that references **module.exports**:
```javascript
// check if IScroll is available in global scope
if (!window.IScroll) {
    // otherwise create local one from module.exports
    var IScroll = module.exports;
}
```

That resolves the issue with CommonJS. Leaving AMD untouched though, unfortunately I don't have opportunity to test AMD as well.


alvarotrigo:
Awesome @AntonyMayer!
Thanks a lot for that! I haven't played much with module bundlers, so I appreciate the help here!

PR #3355, opened 2018-08-08
fix: call wrong object key; +: linear ease; -: console.logs

bpmusicDE:
Checked to use 3.X: window.fp_easings was not defined after run into mobile browser address bar problem with Fullpage 2.X & attached background pics (with option css3: false)

no changes made in minified version of file


alvarotrigo:
Good catch! Thanks for that!
I appreciate it! 👍

bpmusicDE:
No problem :) If the project goes live you'll hear from me for a license

alvarotrigo:
@bpmusicDE awesome to hear! ;)

Issue #3403, opened 2018-09-12
Using autoScrolling:false with scrollOverflow: true conflicting scrollbars


AmanSahota:
Is there anyway to use both `scrollOverflow: true` and `autoScrolling: false` together with `fp-auto-height` within a section without the scrollbars conflicting?

I have a pen here: https://codepen.io/AmanSahota/pen/MqpJzx

The middle section has `fp-auto-height` and I have enabled the browser scrollbar.. however both sections don't seem to work very well together. What would be the best way to do this?

The reason I would like the main scrollbar is because I want to add some animation effects on scroll.

alvarotrigo:
> Is there anyway to use both scrollOverflow: true and autoScrolling: false together with fp-auto-height within a section without the scrollbars conflicting?

Nop. Those options are mutually exclusive. 
In fact, is difficult for me to make any sense of the use of all them together. 

The only purpose of using `fp-auto-height` is to not use `scrollOverflow:true`.  

In any case, I'll have to add it on the warnings function so you'll get an error on the JS console when using both. As it now happens when using `scrollBar:true` and `scrollOverflow:true` as can be seen [here in this line](https://github.com/alvarotrigo/fullPage.js/blob/master/src/fullpage.js#L3134).

alvarotrigo:
This issue stills open for anyone who want to become a contributor to this project! 👍 
I'm giving this way the chance for anyone to contribute to an open source project and practise their Javascript skills.

I propose the solution here. Now you have the chance to work on a few lines and make your contribution! 

Proposed solution
------
On the `displayWarnings` function add an `or` condition in the currently condition that checks for `scrollBar:true` and `scrollOverflow:true`. 
The options `autoScrolling:false` and `scrollOverflow:true` are also exclusive. 

Then on the message replace the following:

> 'Option `scrollBar` is mutually exclusive with `scrollOverflow`.

For 

> 'Options `scrollBar:true` and `autoScrolling:true` are mutually exclusive with `scrollOverflow:true`.

alvarotrigo:
@dotEthan nop. As long as they do not modify the same lines there won't be conflicts.

alvarotrigo:
@dotEthan you can sync or "pull" from the dev branch as I made a few more changes. So you always work on the latest version.

dotEthan:
Updated with pull request #3406

alvarotrigo:
Fixed in the latest version 3.0.3

PR #3406, opened 2018-09-12
Bugfix #3403: autoscrolling and scrolloverflow are exclusive.

dotEthan:
Added an extra option to display a warning the user that autoScrolling and scrollOverflow are exclusive.

### DIFF
--- a/src/fullpage.js
+++ b/src/fullpage.js
@@ -3137,8 +3137,9 @@
                 showError('warn', 'Option `loopTop/loopBottom` is mutually exclusive with `continuousVertical`; `continuousVertical` disabled');
             }
 
-            if(options.scrollBar && options.scrollOverflow){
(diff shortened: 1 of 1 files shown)


alvarotrigo:
Awesome! Thanks for that @dotEthan !
Merged!

alvarotrigo:
This time you forgot to add the issue number (#3403) on the commit title. 
When you do so, the commit appears on the related issue too, which helps to keep track of things.

dotEthan:
Ah, that makes sense, was wondering why it didn't link over this time! Thanks again!

alvarotrigo:
Noticed there was a bug here. It should have been `!autoScrolling`, as there are only problems when using together `scrollOverflow:true` and `autoScrolling:false`.
So I just updated it.

PR #3426, opened 2018-09-27
workaround for issue #44. 

Cai3ra:
Check if WRAPPER_SEL exists before removeListeners on touch events.

### DIFF
--- a/src/fullpage.js
+++ b/src/fullpage.js
@@ -2871,12 +2871,14 @@
                     $body.removeEventListener(events.touchmove, preventBouncing, {passive: false});
                     $body.addEventListener(events.touchmove, preventBouncing, {passive: false});
                 }
+                
+                if ($(WRAPPER_SEL).length > 0) {
+                    $(WRAPPER_SEL)[0].removeEventListener(events.touchstart, touchStartHandler);
+                    $(WRAPPER_SEL)[0].removeEventListener(events.touchmove, touchMoveHandler, {passive: false});
 
-                $(WRAPPER_SEL)[0].removeEventListener(events.touchstart, touchStartHandler);
-                $(WRAPPER_SEL)[0].removeEventListener(events.touchmove, touchMoveHandler, {passive: false});
-
-                $(WRAPPER_SEL)[0].addEventListener(events.touchstart, touchStartHandler);
-                $(WRAPPER_SEL)[0].addEventListener(events.touchmove, touchMoveHandler, {passive: false});
+                    $(WRAPPER_SEL)[0].addEventListener(events.touchstart, touchStartHandler);
+                    $(WRAPPER_SEL)[0].addEventListener(events.touchmove, touchMoveHandler, {passive: false});
+                }
             }
         }
 
@@ -2891,8 +2893,10 @@
                     $body.removeEventListener(events.touchmove, preventBouncing, {passive: false});
                 }
 
-                $(WRAPPER_SEL)[0].removeEventListener(events.touchstart, touchStartHandler);
-                $(WRAPPER_SEL)[0].removeEventListener(events.touchmove, touchMoveHandler, {passive: false});
+                if ($(WRAPPER_SEL).length > 0) {
+                    $(WRAPPER_SEL)[0].removeEventListener(events.touchstart, touchStartHandler);
+                    $(WRAPPER_SEL)[0].removeEventListener(events.touchmove, touchMoveHandler, {passive: false});
+                }
             }
         }
 


alvarotrigo:
Thanks for the PR!
However, I believe this issue should be addressed directly on the vue-fullpage.js component.

It seems that vue-fullpage.js tries to destroy itself when it shouldn't when changing to one route to another one using vue-router. I've tried using `beforeUpdate` or `updated` but those event don't seem to get fired.

I'm a bit lost here, but do you know how could we fire some event before the route changes the DOM?

Cai3ra:
I will try to watch route methods and try to implement a interceptor to the routes.

Get back to you soon.

alvarotrigo:
I think I found the answer!

Using the `beforeRouteLeave (to, from, next) `  [in-component guard](https://router.vuejs.org/guide/advanced/navigation-guards.html#in-component-guards).

alvarotrigo:
That's the one we should be using instead. And only use `beforeDestroy` if `.fullpage-wrapper` is present.

Cai3ra:
I believe that is not going to work because navigation guards works only in route components, when registered in router constructor.

screen shot 2018-09-27 at 14 49 01
screen shot 2018-09-27 at 14 49 53

Cai3ra:
> That's the one we should be using instead. And only use `beforeDestroy` if the element with the class `.fullpage-wrapper` is present.


Do you mean this:
screen shot 2018-09-27 at 14 52 54

alvarotrigo:
> I believe that is not going to work because navigation guards works only in route components, when registered in router constructor.

Mmmm, right...



> Do you mean this:

Yeap, but... we still need to destroy the events attached to the document and window objects, so that won't solve the issue as it won't get destroyed on route change and when initialising fullPage.js again, it will bind the events for the 2nd time.

alvarotrigo:
What we ideally have to do, is only call the destroy method when the whole previous fullpage.js structure still in the DOM.

alvarotrigo:
Uh! I've reverted the merge. 
I realised you didn't commit it to the right branch. You should be pushing it to the `dev` branch.

I've decided to merge it. Fullpage.js shouldn't be firing an error when trying to destroy it if the element doesn't exist.

alvarotrigo:
Can you please create another pull request but pushing it to the dev branch instead?

alvarotrigo:
@Cai3ra let me know if you plan to commit it again. Otherwise I'll do it myself ;)

Cai3ra:
Hi, @alvarotrigo ! Sorry for late. 
I've worked a lot in the last few days.
I'll send the PR again now in the correct branch.
I think of another solution directly in the vue wrapper, if I succeed I send you a new PR. ;)

Thanks.

alvarotrigo:
> I think of another solution directly in the vue wrapper, if I succeed I send you a new PR. ;)

Yeah that'd be nice too! Thanks man!

Although it's true your suggested solution should also be adopted. No component should trigger a console error when trying to destroy the only single instance when it is not even present in the DOM.
That'd be consistent as well with how it works on desktop, where no error is produced.

Cai3ra:
Yes, I think that too.

I will try to work with local injection of component, different of actual implementation in the global vue application and with this I try to have more control of hooks.

alvarotrigo:
Great. I'll be waiting to hear from you then!

Cai3ra:
Ok! Thanks, dude!

Cai3ra:
I can not send a new PR, the branch it is up to date when I try pulling of branch.

I need to send a new PR with this code?

alvarotrigo:
@Cai3ra can't you just create a new pull request to the **dev branch**?
Your changes are clearly not there.
https://github.com/alvarotrigo/fullPage.js/blob/dev/src/fullpage.js#L2896

Cai3ra:
OHH, sorry! I was in my repo that was up to date. I took your changes and merged into my repo.
I sent you a new PR.

alvarotrigo:
@Cai3ra no worries ;)

alvarotrigo:
> I sent you a new PR.

Sorry I though you said "I'll send" :)
I didn't get any PR from you after this one.
https://github.com/alvarotrigo/fullPage.js/pulls

alvarotrigo:
@Cai3ra it seems @splex7 has create [a pull request](https://github.com/alvarotrigo/fullPage.js/pull/3441) that solves the main issue.

I'll wait for a bit more for your commit, as you were the 1st one who proposed it, but otherwise I'll just merge his.

alvarotrigo:
@MerryMacMapface I've merged the the [pull request](https://github.com/alvarotrigo/fullPage.js/pull/3441) on the dev branch. That should solve your issues.

PR #3436, opened 2018-10-05
Bugfix: loss attached events on unwrapping

Khusainov:
There are two bugs with the same root cause that current PR fix:
Assuming you have elements inside your section with attached event listeners (e.g. in case of using angular it can be ng-bind).
1) After initializing FullPage your section content will be wrapped by additional HTML markup.
If you call "destroy('all')" this markup will be removed and content returned to original state. But actually all DOM elements will be recreated, because innerHTML/outerHTML properties are used for unwrapping.
Hence all listeners won't work after that. 
2) Assuming you are using scrolloverflow with FullPage. Your section content will be wrapped by scrolloverflow markup. Now you resize your window to the size when scrollbar should be hidden. As result scrolloverflow markup will be removed. The same method is using for unwrapping section content, hence same bug is here. 
  
Solution:
Created unwrap function that keeps existing DOM objects (instead of recreating them using innerHTML) to save all event listeners attached to them;
Used unwrap function in "fullpage - destroy all" and in "scrolloverflow - destroy".

### DIFF
--- a/src/fullpage.js
+++ b/src/fullpage.js
@@ -3129,7 +3129,7 @@
             [TABLE_CELL_SEL, SLIDES_CONTAINER_SEL,SLIDES_WRAPPER_SEL].forEach(function(selector){
                 $(selector, container).forEach(function(item){
                     //unwrap not being use in case there's no child element inside and its just text
-                    item.outerHTML = item.innerHTML;
+                    unwrap(item);
                 });
             });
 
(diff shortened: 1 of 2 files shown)


alvarotrigo:
Like it!! 👍 
Thanks a lot for that!! I didn't really know that could be done! 

Genius! :)

alvarotrigo:
Btw, perfect PR! 

- Good explanation of the issue
- Good comments for the function (with usage + jsfiddle example!)

PR #3441, opened 2018-10-10
Add conditional statement before remove eventListener

splex7:
1- Make sure to commit it to the 'dev' branch!

In applications that generate and destroy templates dynamically, as like MeteorJS,

templates are destroyed by template engine. 

When fullpage_api.destroy('all') is called, current templates are already removed. 

it makes undefined Object error.

To avoid this, I added just conditional statement $(WRAPPER_SEL)[0] is still exists or not.

(I found this issue when I test laptop device that has touch screen )

Thank you!

### DIFF
--- a/src/fullpage.js
+++ b/src/fullpage.js
@@ -2893,8 +2893,10 @@
                     $body.removeEventListener(events.touchmove, preventBouncing, {passive: false});
                 }
 
-                $(WRAPPER_SEL)[0].removeEventListener(events.touchstart, touchStartHandler);
-                $(WRAPPER_SEL)[0].removeEventListener(events.touchmove, touchMoveHandler, {passive: false});
+                if($(WRAPPER_SEL)[0]){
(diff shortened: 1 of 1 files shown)


alvarotrigo:
Awesome thanks for it!!
Merged!

alvarotrigo:
Also, would you like to commit another PR applying the same condition a few lines up? In the `addTouchHandler` function?

```javascript
if ($(WRAPPER_SEL).length > 0) {
    $(WRAPPER_SEL)[0].removeEventListener(events.touchstart, touchStartHandler);
    $(WRAPPER_SEL)[0].removeEventListener(events.touchmove, touchMoveHandler, {passive: false});
    
    $(WRAPPER_SEL)[0].addEventListener(events.touchstart, touchStartHandler);    
    $(WRAPPER_SEL)[0].addEventListener(events.touchmove, touchMoveHandler, {passive: false});
}
```

PR #3623, opened 2019-04-10
Fixes #3613

### LINKED ISSUE #3613: Multiple Menus
### Description
In the [change from v2.* to v3.*](https://github.com/alvarotrigo/fullPage.js/commit/27f4f8ec22435ce01e73de7ee75395b5d42e9698) there was a regression in the code. In v2.* when providing a selector for the menu options you could provide a class selector and it would target all matched elements, in v3.* it will only target the first matched element.

### Link to isolated reproduction with no external CSS / JS
[Functionality in v2.6.6](https://codepen.io/englishtom/full/OGVoMo)
[Functionality in v3.0.4](https://codepen.io/englishtom/full/pBJZMm)

### Steps to reproduce it
1. Scroll down the page in both provided examples
2. In v2.6.6 both menu's active state are updated
3. In v3.0.4 only the first menu's active state is updated

### Versions 
All browsers
All operating systems
fullpage.js v3.0.4

englishtom:
**Description**
In the change from v2.* to v3.* there was a regression in the code. In v2.* when providing a selector for the menu options you could provide a class selector and it would target all matched elements, in v3.* it will only target the first matched element.

**Link to isolated reproduction with no external CSS / JS**
Functionality in v2.6.6
Functionality in v3.0.4

**This PR Contains**
Passing QUnit tests describing the missing functionality
Updates to `styleMenu` and `activateMenuElement` functions to make them deal with multiple menus rather than just grabbing the first matched element from the given selector

### DIFF
--- a/src/fullpage.js
+++ b/src/fullpage.js
@@ -913,7 +913,11 @@
 
             //moving the menu outside the main container if it is inside (avoid problems with fixed positions when using CSS3 tranforms)
             if(options.menu && options.css3 && closest($(options.menu)[0], WRAPPER_SEL) != null){
-                $body.appendChild($(options.menu)[0]);
+                $(options.menu).forEach(function(domElement, i) {
+                    var menu = $(options.menu)[i];
+
+                    $body.appendChild(menu);
+                });
             }
         }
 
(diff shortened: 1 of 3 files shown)


alvarotrigo:
Looking great! Thanks for that! 👍 
I commented a couple of things in the code.

Can you do those changes I suggested before a couple of hours?
I'm planning to release the next version now :)

englishtom:
Hey Alvaro,

I can definitely take a look at this now, can you point me towards your comments as I don't see any commits?

Thanks,

Tom

alvarotrigo:
Why looking again in the DOM using `$(options.menu)[i]` ? 
Just use `domElement`.

```js
$body.appendChild(domElement);
```

alvarotrigo:
Thanks! I've merged it and modified a couple of variables!

PR #3643, opened 2019-05-12
Clarifies section index starts at 1 and slide at 0

SulphurDioxide:
In reference to alvarotrigo#3626 - clarifying that the first section will have an index of 1 but the first slide (in any section) will have the index of 0.


alvarotrigo:
Merged!
Thanks for that @SulphurDioxide ! 👍

Issue #3330, opened 2018-07-23
Latest scrolloverflow.js update breaks functionality on dev


BReyell:
### Description
On the dev branch, the current scrolloverflow.js breaks functionality of overflow scrolling. Reverting back to the previous version solves the issue.

Throws below error: 
Uncaught TypeError: Cannot read property 'prototype' of undefined

### Steps to reproduce it
Make sure overflow scrolling is active and latest dev branch js files are being used.

To fix: use previous version of overflow.js.

### Versions 
Running Chrome 67 on macOS 10.13.6

alvarotrigo:
Fixed now in fullpage.js 3.0.2

PR #3655, opened 2019-05-31
fix(scrolloverflow): IScroll is not defined when using bundler

gnuletik:
When using scrolloverflow with a bundler like `webpack` or `rollup`, I get the following error

```
IScroll is undefined
```

I import `fullpage` like this: 

```js
await import('fullpage.js/vendors/scrolloverflow')
const { default: fullpage } = await import('fullpage.js')

const myFullpage = new fullpage('.fullpage-container', {
	licenseKey: 'xxxx',
	scrollOverflow: true
});
```

This PR fix the issue

I've seen that there are two related issues.

This PR fixed the issue: https://github.com/alvarotrigo/fullPage.js/pull/3319 but it introduced another bug : https://github.com/alvarotrigo/fullPage.js/issues/3330

This PR fixes both issues.


alvarotrigo:
What's wrong with the way it is done in the webpack example?
https://github.com/alvarotrigo/fullPage.js/blob/master/examples/module_loaders/webpack-example/src/index.js

alvarotrigo:
I'll merge it anyways, as it seems more logical to add the window object there :)

alvarotrigo:
Thanks for that!

gnuletik:
The example has two differences:

- Do not use dynamic import (await import() function):
This is because, I needed to do a dynamic import when my component is mounted in a SSR app to avoid loading fullPage.js on server side (where it fails as windows is not defined).

- Anonymous import (do not get the exported function):
In the example IScroll is imported but not used, so it triggers an error for ESLint.

Thanks for the merge :)

alvarotrigo:
Thanks for the explanation! 👍

PR #3708, opened 2019-08-02
Update Chinese documentation

AnMokoto:
Update Chinese documentation, and replace keywords (section/slide)
thx @alvarotrigo review the pr

1. Make sure to commit it to the `dev` branch!
2. Read https://github.com/alvarotrigo/fullPage.js/wiki/Contributing-to-fullpage.js


alvarotrigo:
Thanks for that!! 
I assume you are Chinese then?

You think it is better to use `section` and `slide` then?

AnMokoto:
Thx fo @alvarotrigo reply. 👍 

We had kept proprietary words such as keywords, methods, parameters, variables, etc. when translating.

`Section/Slide` is the default bearer in HTML structure, and they are an important attributes in `fullpage.js`, ideal to use when to use work while understanding design intent，even may be customize within options `sectionSelector` & `slideSelector` 

Finally  , Im from ShenZhen, a city part of China, Actually I'm Chinese :smile:


PS: `Each section will be defined with an element containing the section class.` & `each slide will be defined by default with an element containing the slide class`

alvarotrigo:
> Finally , Im from ShenZhen, a city part of China, Actually I'm Chinese 😄

Then I'll trust you 100% ! ;)

I've asked someone else and they told me your translation looks easier to read / understand so I'll merge it right now! 👍 

I really appreciate your effort! Thanks a lot!

PR #3750, opened 2019-10-01
swapping var for const

nick-w-nick:


alvarotrigo:
Thanks for that!
But  without using babel or any other transpiler `cons` will throw an error on IE 10 and below.

So I prefer to keep it as `var` :)

PR #3763, opened 2019-10-04
French grammar improvement and website translation added

MercureTony:


alvarotrigo:
That's awesome! Thanks a lot for that @MercureTony! 
May I ask, where did you find about fullPage.js? Was it through http://www.pullrequestroulette.com ?

MercureTony:
It's a pleasure. Actually, I heard about it when I started learning web development around the year 2016 I believe.

alvarotrigo:
Nice! Thanks for the info!

I'll let you know when I get the french translation published in both the site and the docs, but it will probably be on the next release!

PR #3802, opened 2019-11-14
Remove redundancy and hedging from README

haider-ilahi:


alvarotrigo:
Thanks for that! Merged!

PR #3826, opened 2019-12-05
Change transition duration independently

manuelmeister:
On my MacBook I tried to lower the scrollingSpeed, because i wanted the css transition from one section to the next to be faster (at around 250ms). The problem with this was, that with my MacBook (Chrome) the smooth scrolling takes longer than 250ms to come to rest. So then it scrolled over more than one section in one go and it's very difficult to scroll only to the next section.

So I tried to set a shorter transition time via css. But this doesn't work as you can't use silentMoveTo anymore, because the transition time is overriden.

So I came to the conclusion, that the scroll response wait time should be disconnected from the transition time. 

---

P.S. slightly offtopic: I think that 
```CSS
transition: all;
```
**all** is bad for performance, it should be only **transform**. But I don't know if that would break other extensions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -693,7 +693,7 @@ fullpage_api.setScrollingSpeed(700);
 Defines the transition time in milliseconds.
 
 ```javascript
-fullpage_api.setTransitionSpeed(700);
+fullpage_api.setTransitionSpeed(450);
 ```
 ---
 ### destroy(type)
--- a/README.md
+++ b/README.md
@@ -200,6 +200,7 @@ var myFullpage = new fullpage('#fullpage', {
 	//Scrolling
 	css3: true,
 	scrollingSpeed: 700,
+	transitionSpeed: this.scrollingSpeed,
 	autoScrolling: true,
 	fitToSection: true,
 	fitToSectionDelay: 1000,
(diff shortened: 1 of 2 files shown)


alvarotrigo:
Good point! 👍👍
The idea is good! However I would go for a more simple change / approach.

The issue is  basically [here](https://github.com/alvarotrigo/fullPage.js/blob/master/src/fullpage.js#L1879) in the `performMovement` function:

```js
//even when the scrollingSpeed is 0 there's a little delay, which might cause the
//scrollingSpeed to change in case of using silentMoveTo();
if(options.scrollingSpeed){
    clearTimeout(afterSectionLoadsId);
    afterSectionLoadsId = setTimeout(function () {
        afterSectionLoads(v);
    }, options.scrollingSpeed);  /// <--------------- here 
}else{
    afterSectionLoads(v);
}
``` 

So perhaps it will be easier to just change that value with the new option:

```js
 afterSectionLoadsId = setTimeout(function () {
        afterSectionLoads(v);
    }, options.transitionLapse);  /// <--------------- here 
```

A few advantages of this approach:

- Avoiding to modify the option `scrollingSpeed` assures no breaking changes when people update.
- Maintaining the meaning for the value of `scrollingSpeed` also assures the code won't break somewhere else, as it is used in multiple places including the fullpage.js extensions.

We would then have to think if we can apply that to the non CSS3 animation, we probably would have to modify the `scrollTo` function. 

> transition: all;
all is bad for performance, it should be only transform. But I don't know if that would break other extensions.

Yeap, I agree!
However as you said I'll have to check that won't break some extensions.

manuelmeister:
I think your idea is better.
I do not care about non CSS3 animation, because it is even worse for performance than transition:all ;)

alvarotrigo:
Let's reopen it to remind me of it :)

manuelmeister:
Okay :D

manuelmeister:
Is this still relevant? Maybe this would be better as an issue instead of a PR?

alvarotrigo:
> Is this still relevant? Maybe this would be better as an issue instead of a PR?

This was fixed in July 2020. 
I guess I forgot to close the PR :)

PR #3832, opened 2019-12-11
Reassigning value for isWindowFocused on focus event

jbez:
See original commit for more detail.

This adds an eventHandler for focus events to set isWindowFocused to true, otherwise this is left on false forever after a single blur event.

### DIFF
--- a/src/fullpage.js
+++ b/src/fullpage.js
@@ -690,6 +690,9 @@
             //detecting any change on the URL to scroll to the given anchor link
             //(a way to detect back history button as we play with the hashes on the URL)
             window.addEventListener('hashchange', hashChangeHandler);
+            
+            // fixing problems with isWindowFocused never resetting beyond one blur event
+            window.addEventListener('focus', focusHandler);
 
             //when opening a new tab (ctrl + t), `control` won't be pressed when coming back.
             window.addEventListener('blur', blurHandler);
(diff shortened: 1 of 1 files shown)


jbez:
Adding some clarity / a more detailed description and reproduction case:

The page loads, fullpage.js gets initialised via _function initialise()_:

* _isWindowFocused_ gets initially set to true
* _var isWindowFocused = true;_

User triggers a blur event i.e. by clicking into the URL-Bar of the browser:
* _function blurHandler()_ fires, setting _isWindowFocused = false_;

User then returns to the page by focusing it, which currently does not trigger anything.

User now presses the CTRL key for whatever reason, i.e. to CTRL + Page Up scroll.
* _keydownhandler()_ fires, setting _controlPressed = e.ctrlKey;_ (**true**)
* _keyUphandler()_ fires
* * checks variable _isWindowFocused_, which at this point is set to **false**
* * * in turn never reaching the line where _controlPressed = e.ctrlKey_ would reset _controlPressed_ to **false** on the keyUp event.

User now tries to scroll using their mousewheel or trackpad
* _function MouseWheelHandler()_ fires
* * checks for _**!**controlPressed_
* * * _MouseWheelHandler()_ does not do anything for the remainder, until any kind of keyDown event that is not CTRL fires, at which point _controlPressed_ gets reset, but _isWindowFocused_ is still **false** until fullpage.js either gets reinitialised/destroyed or the page is reloaded.

alvarotrigo:
Thanks a lot for that @jbez !! 👍👍👍 
Totally makes sense!

I stopped having access to a windows computer / keyboard long time ago and therefore I am not able to reproduce the issue on my Mac as it  uses a different combination of keys for zooming and so on :)

alvarotrigo:
Merged!

PR #3901, opened 2020-02-25
Add preventDefault option check in scrollOverflow library

luke-m:
Hi @alvarotrigo 

We came across a bug with the scrollOverflow plugin used in fullPage: When scrollOverflow is enabled, mobile links don't work. The event is caught and suppressed. The biggest problem is the `e.preventDefault()`. By removing that, the problem is fixed. Seeing as there is also already an option for this, we used that for this cause. 

Initially we made a fork of the scrollOverflow library but that also didn't work with fullPage because the `scrolloverflow.js` your project uses is modified (the whole `window.fp_scrolloverflow` functionality) so we forked your project instead, making it easier for us and because scrollOverflow doesn't seem to be maintained anymore, it also made more sense to us to fork from your project directly. 

I initially forked from your master branch so there's probably conflicts, but I'm a bit under time pressure so perhaps you can take care of that :) 

Best,
Luke


alvarotrigo:
I can't seem to reproduce your issues regarding clicking on links. 
see the demo here:
https://alvarotrigo.com/fullPage/prueba/examples/scrollOverflow.html

1- Scroll down
2- Click on the huge link which is the 1st part of the paragraph in section 2
3- Click on the yellow part of the paragraph to test JS click events.

luke-m:
Hmm I can't reproduce it on your demo either. One critical information I forgot to mention is that we are using webcomponents and we're thinking that the problem probably stems from the event not being to escape the shadow root or something to that effect. I'll present you a demo of the problem once I have some more time but for now this solution is working for us so we'll just continue using our fork. 
Thanks for the quick reply!

alvarotrigo:
Are you using the latest  version of fullPage.js and scrollOverflow.min.js?
It should be 3.0.8 and Scrolloverflow 2.0.5

alvarotrigo:
I was unable to reproduce the issue in my side.

luke-m:
Hi @alvarotrigo 
Regarding reproducability: I will send you a test-branch of our framework where you can see the issue. Currently, I'm busy with other things but I send you the link as soon as I'm free. 
Also, I will close this PR and open a new one from a different branch because we're currently using the master branch of our fork for further development which has nothing to do with this PR. 
Best,
Luke

alvarotrigo:
> Regarding reproducability: I will send you a test-branch of our framework where you can see the issue. Currently, I'm busy with other things but I send you the link as soon as I'm free.

I just need a jsfiddle or codepen with an isolated reproduction of the issue. Not a whole project :)

PR #3915, opened 2020-03-05
Prevent default option check

luke-m:
Reopened from #3901


alvarotrigo:
Thanks!

However, I'll wait for an isolated reproduction of the issue and a description of what this PR solves before merging it.

Issue #3925, opened 2020-03-13
Mixed Javascript & CSS Vertical Centering of #fp-nav


alanalvarado:
I have been checking the fullpage.js code and I noticed it centers #fp-nav vertically by using javascript combined with css

on the CSS we have

```
#fp-nav {
margin-top: -32px;
top: 50%
}
```
and then in the javascript
```
//centering it vertically
css($(SECTION_NAV_SEL), {'margin-top': '-' + ($(SECTION_NAV_SEL)[0].offsetHeight/2) + 'px'});
```

wouldn't it be better to just use a 2D transform?

```
#fp-nav {
top: 50%
transform: translateY(-50%)
}
```

In this way one would not have to override too much css to move the navigation with callbacks

According to www.caniuse.com 2D transforms are available from ie9

alvarotrigo:
Totally agree :)
Please feel free to open a pull request on the dev branch with the changes and I'll merge it!  👍

alvarotrigo:
Fixed in fullPage.js 3.0.9! 👍

PR #3927, opened 2020-03-14
#3925 Vertical Center #fp-nav Using 2D Transform

alanalvarado:
#3925 Vertical Centering of #fp-nav was achieved by mixing css margin-top
and Javascript which applied directly to the element resulting in
having to use too much css overrides and !important when moving the
navigation with callbacks or styling the vertical navigation menu


alvarotrigo:
Thanks for that!! 👍 
I appreciate it!

Issue #3931, opened 2020-03-20
Error code when using normalScrollElements option and leave mouse point from browser window to debug console.


CressZZ:
### Description
Hi.
I saw follow error code when i use `normalScrollElements` option and leave mouse point from browser window to debug console.
```
scrolloverflow.js:2412 Uncaught TypeError: Cannot read property '0' of null
    at Object.setIscroll (scrolloverflow.js:2412)
    at HTMLDocument.onMouseEnterOrLeave (fullpage.js:766)
```
Actually i think that it may be occurred from `function onMouseEnterOrLeave(e)` in  fullpage.js.

```js
        function onMouseEnterOrLeave(e) {
            // ...

            //onMouseLeave will use the destination target, not the one we are moving away from
            var target = type === 'mouseleave' ? e.toElement || e.relatedTarget : e.target;

            if(target == document || !target){
                setMouseHijack(true);

                if(isUsingScrollOverflow){
                    options.scrollOverflowHandler.setIscroll(target, true);
                }
                return;
            }

            // ...
        }
```
 
Because when i leave mouse point from browser window to debug console, 
1. eventType is `mouseleave`.
2. event.toElement is `null`.

and then  `options.scrollOverflowHandler.setIscroll(target, true);`   in `scrollOverflow.js` is occur error because target is null.
 
`var scrollable = fp_utils.closest(target, SCROLLABLE_SEL) || $(SCROLLABLE_SEL, target)[0];`

```js
 /* scrollOverflow.js */

            // Enables or disables the whole iScroll feature based on the given parameter.
            setIscroll: function(target, enable){
                if(!iscrollHandler.hasBeenInit){
                    return;
                }
                var scrollable = fp_utils.closest(target, SCROLLABLE_SEL) || $(SCROLLABLE_SEL, target)[0];
                var action = enable ? 'enable' : 'disable';
                
                if(scrollable){
                    scrollable.fp_iscrollInstance[action]();
                }
            }

```
So, could you check this matter..

sorry for my writing issue skill. it's my first issue...

Ah!! and i know that this is minor issue to user not developer. they would not see this error. 

### Link to isolated reproduction with no external CSS / JS
[Ideally in jsfiddle.net (https://jsfiddle.net/alvarotrigo/ea17skjr) or codepen.io (https://codepen.io/alvarotrigo/pen/NxyPPp), links to personal websites won't be reviewed unless isolated. Reported issues without a reproduction might get closed.]

[...]

PR #3973, opened 2020-05-11
Fix for #3931 (normalScrollElements option)

AlekseiKrivo:
This is a fix for the issue #3931, that appears when **normalScrollElements** option is added and mouse cursor leaves from a browser window.

When the mouse cursor left the viewport, the **scrollable** variable was not assigned a value, because the **target** parameter, in this case, is null. So, I just added a check for the **target** parameter value in **2409** line.


alvarotrigo:
Thanks for that @AlekseiKrivo !! 👍 
I really appreciate it!

I'll take a look at it for the next release before merging it!

alvarotrigo:
Merged!

PR #3980, opened 2020-05-26
Fixes a small typo in fullpage.js

maxxxxxdlp:
Fixes a small typo in fullpage.js

Replaces `or` with `of` in fullpage.js on the line 3600


alvarotrigo:
Good catch!! 👍 
Thanks for that!

PR #4020, opened 2020-07-31
Patch 1

JayBox325:
I have adjusted how the navigation tooltips appear. Where they were not always central when positioned with `top: -2px` after I'd made it look how I wanted for my design.

I have positioned the tooltip with `top: 50%; transform: translate3d(0, -50%, 0);` so it's always centred with the dot. Bear in mind this includes no cross-browser support in assuming this is done on people's projects with their Sass Compiler.


alvarotrigo:
> Where they were not always central when positioned with top: -2px after I'd made it look how I wanted for my design.

Not always? Under which circumstances?
Can you please provide an isolated reproduction of the issue you are trying to solve?

You can do it on [jsfiddle](https://jsfiddle.net/alvarotrigo/ea17skjr/) or [codepen](https://codepen.io/alvarotrigo/pen/qqabrp), with no CSS or JS files external to fullPage.js and the minimum amount of HTML code? 

Use empty sections unless strictly necessary for the reproduction. issue that this pull request is trying to solve?

JayBox325:
Exciting, thanks!

See here for what my tooltips looked like and note they're aligned to "top" so they're bobbing down mis-aligned: https://jsfiddle.net/g0b5f3dk/

See here for my CSS fix: https://jsfiddle.net/9ayvm24r/2/

alvarotrigo:
> See here for what my tooltips looked like and note they're aligned to "top" so they're bobbing down mis-aligned: https://jsfiddle.net/g0b5f3dk/

I can't see any tooltips there :)

JayBox325:
Here it is again without the top: 50% fix = https://jsfiddle.net/0Lx4mfwt/

al4ever10 (approved):

PR #4488, opened 2022-11-19
docs: fix outdated 'gziped size'

cloydlau:
Hi. The '7Kb gziped' statement seems outdated. Replacing it with a badge so that the number would be always up-to-date.


alvarotrigo:
Awesome, thanks for that! 👍

PR #4489, opened 2022-11-21
docs: make badge row more concordant

cloydlau:
Since there's only one non-badge information left in the badge row, it looks a bit out of alignment and not very concordant:

1aa093082d1f20c4ee7efe7521405ae

We can badge it too (like vercel did in [turbo](https://github.com/vercel/turbo)):

image

It's quite trivial and personal though, I should close this if you don't like it.

This badge can be styled by the way (change background color or remove the twitter logo etc.)


alvarotrigo:
I appreciate it!
It does look good! But I think I'd like to leave this one as plain text as it seems to be it's more visible :) 

Perhaps it's precisely the fact that it doesn't blend as well with the other badges that makes it easier to spot.  

Thanks though! 👍

PR #4556, opened 2023-06-13
Added the <summary> element to focussable elements constant.

JayBox325:
Adding the summary HTML element to the focusable elements constant.

Building an application using the Gov.uk Design System and found a bug where the Details component wasn't focussable.

https://design-system.service.gov.uk/components/details/

1. Make sure to commit it to the `dev` branch!
2. Read https://github.com/alvarotrigo/fullPage.js/wiki/Contributing-to-fullpage.js


alvarotrigo:
Awesome, thanks for that! 
I appreciate it! 👍

PR #4563, opened 2023-06-29
Fix typo: throw->through

jefferyjohn:
Hi! Quick PR to swap 'throw' with 'through' for the description of vertical navigation:

image

Including 'the' before 'sections' also isn't necessary, but I'd consider that a more stylistic choice.


alvarotrigo:
Ups! ;)

Thanks! 👍

PR #4645, opened 2024-08-06
Add Traditional Chinese translated doc

Edit-Mr:
Translated and added Traditional Chinese documentation.


alvarotrigo:
Thanks a lot for that! 👍 
I appreciate it!

Can you add a bit more of context?

Isn't the simplified Chinse enough?
Or isn't it more common on the internet?

Is it possible that some developers know traditional Chinese but not the simplified one?

I'm asking because for us foreigners it's a bit difficult to understand the differences and if adding the traditional Chinese is just a "nice to have" option and not a requirement, perhaps it's not worth the effort to maintain it.

Edit-Mr:
Hey @alvarotrigo ,

The differences between Traditional and Simplified Chinese are not just stylistic; they represent different writing systems used in different regions. For example, people in Taiwan primarily use Traditional Chinese and may find it difficult to read or understand Simplified Chinese. In some cases, there's even a level of aversion or discomfort toward Simplified Chinese due to cultural and political reasons.

While Simplified Chinese is more common on the internet and in mainland China, Traditional Chinese remains the standard in places like Taiwan, Hong Kong, and Macau. Offering Traditional Chinese might be more than just a "nice to have"—it could be crucial for user experience and inclusivity.

I understand that it requires some effort to maintain, but if needed, I’m more than happy to help (:

Edit-Mr:
By the way, the current Simplified Chinese documentation contains quite a few grammatical errors and awkward phrasing, making it seem like it was translated by a machine. Therefore, I translated it from the English documentation. If someone is responsible for this part, they might want to pay attention to it, and if needed, I’m also happy to help.

alvarotrigo:
Right, thanks for the explanation!

Can you update the PR to resolve the conflicts? 
I'll then merge it.

Edit-Mr:
@alvarotrigo Done!

PR #4695, opened 2025-03-31
Remove unused gulp-sass dependency

### LINKED ISSUE #4694: gulp-sass requires Python 2; update to latest version seems to work fine
I tried building fullPage.js from git, and found that it required Python 2, because of its dependency on an old version of gulp-sass (which in turn needs an ancient node-gyp). Simply updating to the latest gulp-sass (6.0.1 at the time of writing) seems to fix that.

rrthomas:
- Build: remove unused dependency gulp-sass (fix #4694)

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,7 +73,6 @@
     "gulp-clean-css": "^4.0.0",
     "gulp-rename": "^1.2.2",
     "gulp-replace": "^1.0.0",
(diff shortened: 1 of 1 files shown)


alvarotrigo:
Thanks for that @rrthomas ! 👍

